PR #5, opened 2018-06-27
Change Swagger JSON to kiali/kiali repo

aljesusg:


theute:
It should have generated a website on netifly, looks like it didn't like the change of owner, I'll have a look.

aljesusg:
mmm It's good to see it now to avoid it in the future with another PRs

PR #9, opened 2018-07-18
KIALI-963 Showcase 'service list' / 'service detail' page

aljesusg:
jul-24-2018 17-22-35

### DIFF
--- a/layouts/features/list.html
+++ b/layouts/features/list.html
@@ -24,13 +24,19 @@
 						<div class="col_three_fifth nobottommargin col_last">
 
 							<div class="heading-block topmargin-sm">
-								<h2>{{ .Title }}</h2>
+
+                                <h2>
+                                    {{ if .Params.file }}
+                                        <a href="/features/{{ .Params.file }}">{{ .Title }}</a>
+                                    {{ else }}
+                                        {{ .Title }}
+                                    {{ end }}
+                                </h2>
 								<span>{{ .Params.subTitle }}</span>
 							</div>
-
 							<p>{{ .Content }}</p>
 
-							<!--a href="#" class="button button-3d button-large">Check out</a-->
+                            <!--a href="#" class="button button-3d button-large">Check out</a-->
 
 						</div>
 
@@ -45,7 +51,13 @@
             <div class="col_three_fifth nobottommargin">
 
               <div class="heading-block topmargin-sm nobottommargin">
-                <h2>{{ .Title }}</h2>
+                  <h2>
+                      {{ if .Params.file }}
+                         <a href="/features/{{ .Params.file }}">{{ .Title }}</a>
+                      {{ else }}
+                         {{ .Title }}
+                      {{ end }}
+                  </h2>
                 <span>{{ .Params.subTitle }}</span>
               </div>
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 output
 public
+
+# IntelliJ
+.idea/
--- a/config.toml
+++ b/config.toml
@@ -3,6 +3,9 @@ languageCode = "en-us"
 title = "Kiali"
 theme = "canvas"
 
+[params]
+    custom_css = ["/css/kiali_style.css"]
+
 [menu]
   [[menu.main]]
   identifier = "swagger"
--- a/content/features/_index.md
+++ b/content/features/_index.md
@@ -1,7 +1,7 @@
 ---
 title: "Features"
 date: 2018-06-20T19:04:38+02:00
-draft: true
+draft: false
 type: "gettingstarted"
 menu:
   main:
--- a/content/features/feature1.md
+++ b/content/features/feature1.md
@@ -1,8 +1,0 @@
----
-title: "Feature 1"
-date: 2018-06-20T19:04:38+02:00
-draft: false
-youTubeID: NcxOapR7D0I
-subTitle: "It rocks"
----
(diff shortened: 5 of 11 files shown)


content/features/service_discovery.md
```diff
@@ -0,0 +1,16 @@
...
+image: "features/service_discovery.png"
+subTitle: "Service view, Pods, Deployments, Rules, Metrics..."
+description: "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Curabitur rutrum pharetra tristique. Vivamus nisl ipsum, interdum at accumsan nec, molestie sit amet dolor. Sed mattis eros leo, ut sagittis mauris fermentum eget. Donec ut mauris interdum, accumsan lacus id, dignissim tortor. Maecenas non risus eget dolor interdum convallis tincidunt efficitur. Etiam accumsan justo sed enim vestibulum, vel ultrices nisl pellentesque. Vestibulum quis est ex. Praesent luctus dapibus mi non commodo. Praesent in tortor euismod, finibus nulla ut, dapibus lacus. Suspendisse ac odio elit. Praesent quis ultricies justo. Vestibulum eu dapibus arcu, quis faucibus ex. Aliquam sit amet suscipit libero. Cras tempor varius orci, ac auctor nibh efficitur id."
+---
+
+# sada
+
+<center>
+  <iframe width="360" height="215" src="https://www.youtube.com/embed/oG-WDA-hmvA" frameborder="0" allow="autoplay; encrypted-media" allowfullscreen></iframe>
+</center>
```
theute: I would put this as the single.html template, similar to https://github.com/kiali/kiali.io/blob/master/layouts/api/single.html it  would be in layouts/features/single.html and applied to all single page feature. Rather than copy/pasting the inclusion code for YouTube, and repeat the YouTubeID

theute: PS: I am not sure if we need a whole page to describe a feature. Explaining a new feature should remain "lightweight" or I'm afraid we won't do / update them as often as needed

lucasponce:
Side comment: I guess the right JIRA is this https://issues.jboss.org/browse/KIALI-1137
The KIALI-963 is the parent one and can be reused for graph or other subtasks.

content/features/service_discovery.md
```diff
@@ -0,0 +1,22 @@
+---
+title: "Service Discovery"
+file: "service_discovery"
+date: 2018-07-18T12:33:50+02:00
+draft: false
+youTubeID: oG-WDA-hmvA
+subTitle: "Service view, Pods, Deployments, Rules, Metrics..."
```
lucasponce: The "..." sounds weird here.
What about ?
"Services view focus on Istio observability"

Just to show something generic, for me "Pods/Deployments/Rules/Metrics" is more a detail it can go below.

aljesusg: Right

content/features/service_discovery.md
```diff
@@ -0,0 +1,22 @@
+---
+title: "Service Discovery"
+file: "service_discovery"
+date: 2018-07-18T12:33:50+02:00
+draft: false
+youTubeID: oG-WDA-hmvA
+subTitle: "Services view focus on Istio observability"
+---
+* List of Services with Health status and error rate
```
lucasponce: I am not sure what's the idea of @theute here, but perhaps we can just summarize in three lines the features.
In order to not need to click on the "more".

Perhaps in three lines:
- List of Services focus on Observability (Health, Error rate).
- Browse and validate Istio networking configuration per Service.
- Detailed Metrics and Tracing.

Or something like that, just an idea. I guess in the video it is showed more or less with a lot of details that perhaps don't need to be listed on the page but just to add some highlights.

aljesusg: I don't know, maybe the Content will have a images and more description about how to do it things in next releases so I think that in the list of features we should make a resume or something like that and show all the information in the full page of the feature

theute:
I think that in general it should fit in few lines and short videos (and rather split in various features), so few comments:
  - Make the 250 longer, I think we can put all your text in this example, there might be other situations where we need a much longer one
  - Looks like there are issues in the bullets for service discovery
  - Let's align the filenames/URLs one is using '-' and the other is using '_', I personally prefer '-' (like on medium.com) but can live either way as long as it is consistent :)
  - You can change to draft=false in features page so that it appears in the menu

Thanks !

layouts/features/list.html
```diff
@@ -24,13 +24,19 @@ <h1>Features</h1>
...
+                                <h2>
+                                    {{ if .Params.file }}
+                                        <a href="/features/{{ .Params.file }}">{{ .Title }}</a>
+                                    {{ else }}
+                                        {{ .Title }}
+                                    {{ end }}
+                                </h2>
 								<span>{{ .Params.subTitle }}</span>
 							</div>
+							<p>{{ .Content }}}</p>
```
theute: There are 3 "}" instead of 2

theute:
For the bullet alignement it seems to be because of the theme style:
dl, dt, dd, ol, ul, li {
	margin: 0;
	padding: 0;
}

In kiali_style.css try to override with:

ul {
  padding-left: 15px
}

I tried quickly it didn't seem to break anything else...

aljesusg:
I read something about configuration file to set a tabulation parameter, it seems that the problem is here because bullets are different.

Anyway Hugo 0.47 is out @theute with ref feature. Should we update this?

theute:
I'm not sure how it's going to help, but we already use the latest AFAIK:

https://github.com/kiali/kiali.io/blob/8ccae4031bd349371e83394961cb067aabf90d7e/.travis.yml#L5

aljesusg:
Upgrade Hugo not help ^^ but I think that we should be in the last version

aljesusg:
### Final Result
captura de pantalla 2018-07-23 a las 10 51 12


To avoid the space between elements of list
```css
ul>li>p{
  all: unset;
}
```
it seems that Hugo keep elements inside of p element
captura de pantalla 2018-07-23 a las 10 51 23

Review this please @theute

aljesusg:
Mmm ok, let me check how we can format only the features list.

aljesusg:
Sorry by delay I was in a fly ^^^
@theute Fixed css
jul-24-2018 17-22-35

PR #12, opened 2018-07-31
KIALI-1086: add graph features

mwringe:
Initial commit for adding in the graph features.

Using images instead of screencasts. Screencasts are great to demo things, but they are difficult to keep up to date with changes. Its much easier to just put in a new screenshot if something changes than to redo a whole screencast.

[I also personally don't want to watch a 5 minute video just to see what features are available or not, its much quicker to read and see some images to get an idea of what it can do]


lucasponce:
Screenshots are good.
My only comment is the consistency and I guess @theute should define here the structure of the content.
AFAIK the video @aljesusg created was a suggested and I also created a video following the previous content.
So, I don't really matter to create a video or screenshots, but at least all content should be similar.
(Indeed, technically speaking screenshots are easier).

aljesusg:
In my opinion it would be great have screenshots for each featuresl and a full video like tutorial in each major version released.

theute:
Features list was designed with either a pic or a video, sometimes a pic or animated gif is enough sometimes not...

PR #13, opened 2018-08-08
Use always https for font import

xeviknal:
Accessing to kiali.io in https triggers an error of insecure request because of the font import.
Changing this import to https makes kiali works for both http and https since it is not a problem (the opposite) to make request to https from a non-secure webpages.


abonas:
@xeviknal how can your fix be tested locally?

xeviknal:
@abonas I only could test the HTTP scenario in local. However, I did try the HTTPS scenario using https://kiali.io production webpage. I went down to the link html node and switch the HTTP schema.

screen recording 3

In the above gif, you can see that adding the https back triggers the non-secure connection error.

PR #27, opened 2018-09-05
Fix the readme index

aljesusg:


aljesusg:
Here is how looks like https://github.com/kiali/kiali.io/blob/e6731408a7c16ba63b3278ba17d042b5cecda475/README.adoc

PR #29, opened 2018-09-05
KIALI-1481 Force to use a specific version in Hugo

aljesusg:


lucasponce:
Instead to commit into the repo the hugo dependency.
Cannot be pulled with explicit version by travis or some script ?
Or what is the main motivation ? I understand we want to force a version, and depending of the rpm deployed we can get one or another, but if it is not complicated, I would try to script the installation steps for each machine.

aljesusg:
Checking why this fail in travis https://travis-ci.org/aljesusg/kiali.io/builds/424740583?utm_source=github_status&utm_medium=notification please **DO NOT MERGE**

PR #20, opened 2018-09-04
Doc overview: fix several typos / errors

jotak:
+ Some small additions / reformulations

PR #31, opened 2018-09-06
Doc overview: more feedbacks

jotak:
Integrate feedbacks from @JStickler here https://github.com/kiali/kiali.io/pull/20#pullrequestreview-152624142


JStickler:
@jotak Thanks for following up on this!  Sorry for the late feedback after the merge, I think just had that issue open in my browser too long.  =P

PR #33, opened 2018-09-07
KIALI-1426 Extend graph issues description and add graph expectations

israel-hdez:


mwringe:
Its a bit long and I don't think we need all the screenshots. I think this should be more short and to the point. 

I think I liked what we had before better:

"The layout for Kiali Graph may render differently, depending on the data to display (number of graph nodes and their interactions) and it's sometimes difficult, not to say impossible, to have a single layout that renders nicely in every situations. That's why Kiali offers a choice of several layout algorithms. However, we may still find some scenarios where none of the proposed algorithm offer a satisfying display."

And then maybe add something at the end:
"If Kiali doesn't not render your graph layout in a satisfactory manner please switch to another layout option. This can be done from 'Graph Settings' by changing the 'Layout Schema' [maybe include a screenshot]

I do like the part you added at the end:
"If Kiali doesn't produce a satisfactory graph for you, don't hesitate to https://github.com/kiali/kiali/issues/new?template=bug_report.md[open an issue in GitHub] or reach us in the https://groups.google.com/forum/#!forum/kiali-users[mailing lists], and describe your situation."

israel-hdez:
@mwringe Ok, good. I'll summarize the text again. Perhaps may be worth to keep a shorter version of the "guidelines" part? I though that part was the one covering the "graph layout expectations".

mwringe:
@israel-hdez We might want to put the guidelines somewhere else, perhaps in the in-app documentation. I don't know if putting it in the known issues section is the right place or not.

israel-hdez:
@mwringe Ok, I reverted the text and added your suggested paragraph.

I'm no longer referring to the Jira in this PR, since it will remain unsolved.

PR #41, opened 2018-10-02
KIALI-1659 Document running with reduced permissions

mwringe:
Documents how to run Kiali with reduced permission in OpenShift.

### DIFF
--- a/content/gettingstarted/index.adoc
+++ b/content/gettingstarted/index.adoc
@@ -159,3 +159,27 @@ external_services:
   ...
 ----
 {nbsp} +
+
+=== Reducing Permissions in OpenShift
+
+By default Kiali will run with the cluster wide `kiali` role.
+
+If you prefer not to run Kiali with cluster wide permissions, it is possible to reduce these permissions to individual namespaces.
+
+icon:lightbulb[size=2x] {nbsp}{nbsp}{nbsp}{nbsp} This only works for OpenShift since it can return a list of namespaces that a user has access to. Know how to make this work with Kubernetes? Awesome, please let us know in this https://issues.jboss.org/browse/KIALI-1675[issue].
+
+The first thing you will need to do is to remove the cluster-wide permissions that are granted to Kiali by default:
+
+[source,bash]
+----
+oc delete clusterrolebindings kiali
+----
+{nbsp} +
+
+Then you will need to grant the `kiali` role in the namespace of your choosing:
+
+[source,bash]
+----
+oc adm policy add-role-to-user kiali system:serviceaccount:istio-system:kiali -n ${NAMESPACE}
+----
+{nbsp} +


content/gettingstarted/index.adoc
```diff
@@ -159,3 +159,27 @@ external_services:
   ...
 ----
 {nbsp} +
+
+=== Reducing Permissions in OpenShift
+
+By default Kiali will run with the cluster wide `kiali` role.
+
+If you do not want to run Kiali with cluster wide permissions its possible to reduce these permissions to individual namespaces.
```
abonas: I suggest: "If you prefer not to run Kiali with cluster wide permissions, it is possible to reduce these permissions to individual namespaces."

content/gettingstarted/index.adoc
```diff
@@ -159,3 +159,27 @@ external_services:
...
+----
+oc delete clusterrolebindings kiali
+----
+{nbsp} +
+
+Then you will need to grant the `kiali` role in the namespace of your choosing:
+
+[source,bash]
+----
+oc adm policy add-role-to-user kiali system:serviceaccount:istio-system:kiali -n ${NAMESPACE}
```
abonas: what is istio-system representing in this example?
is there any requirement for specific namespaces to have this in order that Kiali works ok? (for example - the namespace where Kiali is deployed, the namespace where Istio is deployed which might be different from the one where Kiali resides, etc.)

mwringe: >what is istio-system representing in this example?

"system:serviceaccount:istio-system:kiali" means its the service account named 'kiali' in the 'istio-system' namespace. Since this is where Kiali is installed, users will always have to use this value.

> is there any requirement for specific namespaces to have this in order that Kiali works ok? 

Nope. If you remove all the namespaces you can't do much in Kiali since everything is empty. But there is no requirement for any particular namespace to be added here.

content/gettingstarted/index.adoc
mwringe: Updated.

content/gettingstarted/index.adoc
abonas: I think I'm missing something here, but if you are granting a role to an account in istio-system namespace, why should a user specify also a namespace in addition to that via the -n flag?

mwringe: The Kiali pod is installed in the istio-system namespace and its running with the service account named 'kiali'. In this case its 'username' is __system:serviceaccount:istio-system:kiali__ 

Kiali needs access to kubernetes resources from within the cluster. Is it to be able to do things such as read pod or deployment definitions. Required to do things such as health checks and checking if a side car is installed.

We have a role called __kiali__ that includes all the resource that kiali needs to have access to and what actions can be performed on them (eg the ability to read pod definitions)

Roles can be granted at the namespace level (eg Kiali can read all pod definitions in namespace 'foo') or a the cluster level (eg Kiali can read all pod definitions for all the namespaces in the cluster)

Going back to this command:

> oc adm policy add-role-to-user kiali system:serviceaccount:istio-system:kiali -n ${NAMESPACE}

Another way to write this could be:

> oc adm policy add-role-to-user ${ROLE_NAME} ${USERNAME} -n ${NAMESPACE}

This means to grant the user named ${USERNAME} the role named ${ROLE_NAME} in the namespace named ${NAMESPACE}.

If the role includes the ability to read pod definitions, then the user specified will be able to read the pod definitions in the specified namespace.

So what this means is to grant the Kiali pod (eg __system:serviceaccount:istio-system:kiali__) the permissions it needs (eg the __kiali__ role) in a specified namespace:

> oc adm policy add-role-to-user kiali system:serviceaccount:istio-system:kiali -n ${NAMESPACE}

mwringe: @abonas Is there anything else that needs to be done or can this be merged now?

abonas:
LGTM

PR #48, opened 2018-10-17
Simplify hugo install

josejulio:
Instead of downloading the code and compiling, download the binary
  and put it on ${GOPATH}/bin

I did this because I was getting dependencies error when trying to build that specific hugo version, probably related to my env, but this seems safer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,10 +3,8 @@ HUGO_VERSION ?= 0.48
 
 install:
 	@echo Installing Hugo version ${HUGO_VERSION}
-	@mkdir -p ${GOPATH}/src/github.com/gohugoio/
-	@wget -qO- https://github.com/gohugoio/hugo/archive/v${HUGO_VERSION}.tar.gz | tar xvz -C ${GOPATH}/src/github.com/gohugoio/
-	@mv ${GOPATH}/src/github.com/gohugoio/hugo-${HUGO_VERSION} ${GOPATH}/src/github.com/gohugoio/hugo; cd ${GOPATH}/src/github.com/gohugoio/hugo; go build -o ${GOPATH}/bin/hugo
-	@export PATH=${PATH}:${GOPATH}/bin
+	@HUGO_VERSION=${HUGO_VERSION} sh scripts/get-hugo.sh
+	@export PATH="${PATH}:${GOPATH}/bin"
 	@echo Installing AsciiDoctor
 	@gem install bundler
 	@bundle install
@@ -23,5 +21,3 @@ deploy: build
 	aws s3 sync public/ s3://kiali.io --acl public-read --delete
 	aws configure set preview.cloudfront true
 	aws cloudfront create-invalidation --distribution-id E3RFW0PBPFCVRO --paths '/*'
-
-
--- a/README.adoc
+++ b/README.adoc
@@ -32,6 +32,19 @@ export PATH=${PATH}:${GOPATH}/bin
 make install
 ----
 
+This will attempt to determine the binary of Hugo needed and download it. If it can't it will proceed to download the source code and compile.
+You can force the compile by setting `BUILD_HUGO=TRUE`
+
+[source, bash]
+----
+BUILD_HUGO=TRUE make install
+----
+
+You can always download the required dependencies yourself if the above fails:
+
+ - Hugo at least v0.48 ([github releases](https://github.com/gohugoio/hugo/releases/tag/v0.48) or brew if you are on mac)
+ - AsciiDoctor (by running `bundle install`)
+
 ==  Start local server
 
 Hugo has a live `serve` command that runs a small, lightweight web server on your computer so you can test your site locally without needing to upload it anywhere.  As you make changes to files in your project, it will rebuild your project and reload the browser for you.
--- a/scripts/get-hugo.sh
+++ b/scripts/get-hugo.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+function arq {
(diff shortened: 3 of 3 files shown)


josejulio:
> The point of download de source code is for Mac users

ooh! Got it. I'll try to see if can do some quick env detection and fallback to "download code and compile"

cfcosta:
Why can't we build it on Docker? I mean, all of us have it installed and configured already.

josejulio:
> Why can't we build it on Docker? I mean, all of us have it installed and configured already.

You mean running "hugo" inside docker? or building hugo binary on docker and then pulling it out?
If the later, it would be the same problem with OSX i guess.

pilhuhn:
As a Max user I prefer `brew install hugo`. 
And tbh if we need huge scripts and compile orgys to get hugo to run and hugo is so very version dependent, then I doubt that this is the right tool for the job.

josejulio:
> As a Mac user I prefer `brew install hugo`.

This will only install hugo for the specific `GOPATH` you set, but i get your point, I also prefer `brew` to install software on my mac :)

> and hugo is so very version dependent

I think we need hugo from 0.48 and upwards because of some feature (not sure which) and the version in fedora older than that.

aljesusg:
Yep @josejulio fedora 27 not support versions older than 0.38 and throws warnings while fedora 28 support 0.48, so we decide install Hugo from source code by that. We need to work with the same version to avoid this things. I am worried about which version install brew.

pilhuhn:
Brew installs 0.48. Usually it is pretty up to date.

aljesusg:
Great so maybe we can install with brew when OSX. I'll dismiss review by PTO

josejulio:
Updated the README and the Makefile to work on Mac

mtho11:
Why don't we use some hosted solution that doesn't require all this nonsense?
Especially considering we want external contributors to actually contribute.

cfcosta:
@mtho11 jekyll?

@josejulio @pilhuhn tbh I'd still prefer to have everything on docker, mainly because we already require docker for EVERYTHING, might as well not have ambient dependencies, just pull an image and build/serve from there.

mtho11:
@cfcosta anything that is hosted and doesn't require me to install and setup stuff to do documentation. I just want to go to a website and start editing and then submit my changes for approval.

mtho11:
Maybe @JStickler has seen some good tools?

pilhuhn:
@cfcosta I am not against Hugo + adoc per se, as this ties nicely in our workflows.
My pain point is once you have to compile Hugo yourself to get the exactly right point version. And/or when we start writing scripts to try to get this right. 
Using it via Docker or homebrew is fine. 
But everything else distracts us from what we really want to do.

JStickler:
Hugo is a LOT less version dependent than Jekyll, which is why I've seen several projects move off Jekyll and onto Hugo.

josejulio:
> My pain point is once you have to compile Hugo yourself

You no longer require to compile hugo yourself with this, `make install` will take care of downloading the right binary for you (on most cases).

Anyway I updated the readme to let the users know where they can download hugo binary if they want.

mtho11:
@JStickler I don't care for Jekyll either. I was wondering if any other groups are using a self-hosted or SaaS style documentation tool? A no tooling needed solution. Something where you just go to the site and write your documentation (no need to run or setup anything locally).

mtho11:
I actually liked GitBook when we used it for Hawkular.

JStickler:
@mtho11 I haven't really looked into it that deeply, there might be something out there?  Don't know.

jmazzitelli:
hugo is awesome and makes creating/proofreading/reviewing doc PRs easy - and it is the same thing Istio uses for their docs. In fact, Istio also has a way that when a doc PR is merged to master, a preliminary doc server is deployed so you can see the current state of the docs for the master branch: https://preliminary.istio.io/docs/ (This is what the 1.1 docs would look like if they were released today.)

pilhuhn:
@jmazzitelli Can you find out how they do that and if we can easily copy&paste what/how they do it?

jmazzitelli:
@pilhuhn I can tell you they use Netlify: https://github.com/istio/istio.io/blob/master/README.md#how-versioning-works

They created an Istio account on Netlify and use the github integration with netlify to do the publishing: https://www.netlify.com

josejulio:
Unless i'm misunderstanding, we already have our workflow to deploy: https://github.com/kiali/kiali.io/blob/master/Makefile#L22

Once a change goes into master it goes directly to kiali.io (istio does a deploy to preliminary instead).

theute:
As a note, I never used the Makefile (it's used by and for Travis).
I do:
 sudo dnf install hugo
 sudo dnf install asciidoctor
 hugo serve

and I'm done...

v0.37.1 is the version that I get on my Fedora 29 and I don't know what I'm missing beside the swagger doc.

I think those should be the instructions in the README and leave the Makefile (with binary download) for Travis

Don't the "same" work on Mac ?  ie: use brew for hugo + asciidoctor ?

cfcosta:
> If we want to make that we will need to start making "releases" of this repo too, and have each branch go to vthe-branch.kiali.io, having our "last" release also on kiali.io and master to preliminary.kialio.io

I like this, tbh. It would require us to also tag the website on release, but I like the idea of having every PR being deployed somewhere for evaluation.

theute:
> but I like the idea of having every PR being deployed somewhere for evaluation.
It is already the case...
See on this PR the "Details" link:
image

josejulio:
#72 seems a better choice.

PR #50, opened 2018-10-18
Update menu-content height

josejulio:
Take into account the top offset introduced

I just noticed this, when the screen becomes small (or the menu bigger due the number of elements) 60px of the bottom will hide and we won't be able to see them because of the offset introduced.

Without fix:
without

With fix:
with


abonas:
is this about a specific section in the website?

josejulio:
Is related to the new API section (see this #47)

PR #52, opened 2018-10-25
JIRA-1747 Update front page background image and colors

abonas:
website

and the second part of the page 

website2


abonas:
@lucasponce absolutely, there are many more changes in queue for the website. (font changes, feature visuals, etc.).
I prefer to send separate PRs on screenshots, features, fonts.

PR #57, opened 2018-11-06
Fix Kiali description

abonas:
Match it to be the same as the more correct description that is in the source code readme.

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -28,7 +28,7 @@
         <div class="col-md-5">
           <div class="slider-title">
             <h2 class="text-white text-rotater mb-3" data-separator="," data-rotate="fadeIn" data-speed="3500">Service mesh observability.</h2>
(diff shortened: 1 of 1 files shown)


pilhuhn:
For this you don't need Kiali and the static graph thingy would be enough.

theute:
I agree with Heiko, this only defines the graph, we need a better description and I agree it should match the Readme.
Maybe something as dull as "Observability console and service mesh configuration for Istio" would be clear enough for the targeted audience... Or something along those line... we can start a thread on one of the mailing lists.

abonas:
I propose to merge it, as this is anyhow more accurate than the current description that is misleading about what microservices are doing.
And absolutely kick another duscussion about improving it.

abonas:
@theute @pilhuhn ^^ ?

theute:
Ok, but please start the discussion or it will be forgotten

PR #59, opened 2018-11-15
Add greek origin explanation

abonas:
addgreekexplanation


abonas:
Also approved by Dimitris A. from greek language perspective.

PR #60, opened 2018-11-18
Update font type and size for title, description and nav menu

abonas:
after:
fontchange


theute:
It looks a lot better !

abonas:
> It looks a lot better !

@theute I'm glad to hear. We experimented with like 4 types of fonts :)
More changes to come soon

PR #67, opened 2018-11-20
KIALI-1965 Use a docker image to start Hugo instead of building

mwringe:
This will use a Hugo docker image to run Hugo instead of having us download and build a version of Hugo.

This simplifies the build process a lot and makes it easier to get things up and running.

Other than requiring docker to be installed, this should behave the same as before.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,18 +1,17 @@
 
 HUGO_VERSION ?= 0.48
+# The port and ip that the Hugo server will bind to
+BOUND_IP ?= 127.0.0.1
+PORT ?= 1313
 
 install:
-	@echo Installing Hugo version ${HUGO_VERSION}
-	@mkdir -p ${GOPATH}/src/github.com/gohugoio/
-	@wget -qO- https://github.com/gohugoio/hugo/archive/v${HUGO_VERSION}.tar.gz | tar xvz -C ${GOPATH}/src/github.com/gohugoio/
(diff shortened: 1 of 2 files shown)


josejulio:
`make build` still calls `hugo` directly. Also, does this work in travis too? https://github.com/kiali/kiali.io/blob/master/.travis.yml#L10

mwringe:
hmm, your right. And its as simple to just use the existing docker image for these since the we need to install asciidoctor. This needs some more work.

theute:
But sorry to ask again but what's wrong with:
  - sudo dnf install hugo
  - sudo dnf install asciidoctor 
?
(and equivalent on Windows or Mac)

Or at worse get the binary for a specific version and add "hugo" in your path

We may use different versions, but I haven't seen this been a problem in practice, or did I miss it ?

abonas:
will this mode pick changes from local file system when running inside a container? e.g. changing css, changing html - will this be picked on the fly like today?

mwringe:
> But sorry to ask again but what's wrong with ... Or at worse get the binary for a specific version and add "hugo" in your path

Installing with DNF has the issue where it might bring in another version than what we are running, so what might work on a users machine might not work in the live version. 

But just saying install Hugo and make sure its on your path is probably by far the easiest solution to all of this.

mwringe:
> will this mode pick changes from local file system when running inside a container? e.g. changing css, changing html - will this be picked on the fly like today?

Yes. We mount the local directory into the docker image and it works exactly like if Hugo as just running locally in that directory.

mwringe:
Closing this PR as it looks like we should make something more simple and just ask users to install Hugo directly.

PR #68, opened 2018-11-21
Replace hero background svg with smaller file

abonas:


theute:
Much lighter/faster ! :)

PR #70, opened 2018-11-21
Remove animation

abonas:


theute:
It definitely seems faster this way, thanks !

abonas:
@theute it was your suggested solution :)

PR #73, opened 2018-11-28
KIALI-2011 fix strange hover in documentation link

abonas:
After the fix the link color is changed, but the size remains the same - which fixes the strange hover issue.
afterfix

### DIFF
--- a/static/css/kiali_style.css
+++ b/static/css/kiali_style.css
@@ -28,11 +28,17 @@ p {
   background-image:url("/images/hero-background.svg")
 }
 
-#primary-menu ul li > a {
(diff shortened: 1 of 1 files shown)


theute:
While it fixer the issue, now the top menu font size changed and going to Documentation>Glossary>Observability, the menu items grow in size when hovered

abonas:
> While it fixer the issue, now the top menu font size changed and going to Documentation>Glossary>Observability, the menu items grow in size when hovered

I changed the top menu font size to be bigger anyway, the Documentation->Glossary->Observability was not working prior to the fix as well, but I'll see how to fix that one too.

abonas:
@theute the issue with secondary menu size is now solved

mtho11:
I'll let @theute push the green button if he approves

PR #82, opened 2018-12-19
Increase mailing list envelope icon size and make it clickable

abonas:


layouts/partials/footer.html
```diff
@@ -32,7 +32,7 @@
 
 
 
-								<i class="icon-envelope2"></i> <a href="https://groups.google.com/forum/#!forum/kiali-users">Kiali Users group</a> <span class="middot">&middot;</span> <i class="icon-envelope2"></i> <a href="https://groups.google.com/forum/#!forum/kiali-dev">Kiali Developers group
+								<a href="https://groups.google.com/forum/#!forum/kiali-users"><i class="icon-envelope2"></i> Kiali Users group</a> <span class="middot">&middot;</span> <a href="https://groups.google.com/forum/#!forum/kiali-dev"><i class="icon-envelope2"></i> Kiali Developers group
```
aljesusg: Missing the close ```</a>``` at the end of the line, it seems that was missed before this change :( Can you add it?
The problem with this is that if we add the ```</a>``` we'll see the license in black because it was using the same css by the ```<a>```

after

So my proposal is add this part to the same link because is related with the license
```html
<div class="widget clearfix">
  <p>
    <a href="https://www.apache.org/licenses/LICENSE-2.0.txt">
      <strong>License:</strong> 
      Kiali is open source and released under Apache License v2
    </a>.
  </p>
</div>
```

static/css/kiali_style.css
```diff
@@ -14,6 +14,7 @@
 
 a {
   color: #0094DE;
+  font-size: 16px;
 }
```
aljesusg: About this I'm worried about this apply to all the links. What do you think about to create a new class for them with a hover effect for this link ?
```html
<a href="https://groups.google.com/forum/#!forum/kiali-users" class="email-group">
  <i class="icon-envelope2"></i>
  Kiali Users group
</a>
<span class="middot">&middot;</span>
<a href="https://groups.google.com/forum/#!forum/kiali-dev" class="email-group">
  <i class="icon-envelope2"></i> Kiali Developers group
</a>
```
```css
.email-group {
  font-size: 16px;
  opacity: 0.8;
}

.email-group:hover {
  opacity: 1;
}
```
This is the demo

a hover

mtho11:
@abonas can you also align those icons to match the others to address: https://issues.jboss.org/browse/KIALI-2174

abonas:
> @abonas can you also align those icons to match the others to address: https://issues.jboss.org/browse/KIALI-2174

please check the preview as deployed in netlify, I believe now it is much more aligned. wdyt?

PR #85, opened 2018-12-20
KIALI-2178 Submenu documentation is white when we hide the header

aljesusg:
### Before

peek 2018-12-20 11-25

### After

peek 2018-12-20 12-14

### DIFF
--- a/static/css/kiali_style.css
+++ b/static/css/kiali_style.css
@@ -39,6 +39,10 @@ h2 {
   color: #FFFFFF;
 }
 
+#primary-menu ul ul li > a {
+  color: black;
+}
+
 #content .features-bottom-box li > a {
   font-size: 16px;
   font-weight: 500


static/css/kiali_style.css
```diff
@@ -39,6 +39,10 @@ h2 {
   color: #FFFFFF;
 }
 
+#primary-menu ul ul li > a {
+  color: black;
+}
+
```
aljesusg: This overwrite the color https://github.com/kiali/kiali.io/blob/2dafe44ce5d77fed6dd2a04cc39d535dc82f8195/static/css/kiali_style.css#L35-L40

it's necessary because we need to apply the rest of css rules

aljesusg:
About the design, Should I apply the rules for the both cases @abonas ? (With header and not header)

abonas:
> About the design, Should I apply the rules for the both cases @abonas ? (With header and not header)

I think so. there was not anything about it in the design, but it makes sense to me. wdyt?

jmazzitelli:
Why is the background white for the menu when the header is hidden? I think it would make more sense for the menu background to be black always, whether the header is hidden or not.

abonas:
@jmazzitelli it should not be not white and not black :)
see above mentioned JIRA to update it to the relevant blue shades.

jmazzitelli:
> see above mentioned JIRA to update it to the relevant blue shades.

Oh.. OK... that would be good too :) Should we close this PR then?

aljesusg:
I am working in this PR to apply the styles of the design document @jmazzitelli, I opened the JIRA because now there is a bug in kiali.io , you can see the menu with the background white when scroll down...I can open another PR to do the JIRA that @abonas said.

aljesusg:
I saw the proposal by design and the changes that we need to make to the css and js of the canvas theme. I'll close this PR and I'll open another one to make this changes because i think that will be cleaner create a theme for Kiali.

aljesusg:
What do you think about merge this while I work in the new styles @abonas @jmazzitelli ? Because now the user can't see the submenus when he scroll down in kiali.io

PR #89, opened 2019-01-08
KIALI-2142 Update styling on website drop down menu of Documentation section

aljesusg:
Remove the dark class we keep the background style of the dropdown in white.
peek 2019-01-08 12-38

I am worried about this. 

The javascript file [themes/canvas/static/js/functions.js](https://github.com/kiali/kiali.io/blob/master/themes/canvas/static/js/functions.js) of the canvas theme applied the style [dark.css](https://github.com/kiali/kiali.io/blob/master/themes/canvas/static/css/dark.css) when the user scroll down, this file make all the magic adding and removing classes. 

Right now we are overwriting the css rules in the [kiali_style.css](https://github.com/kiali/kiali.io/blob/master/static/css/kiali_style.css) file in some cases with **!important**

We are changing  practically the canvas theme and I think that we should create a new theme for Kiali (Jaeger website has their own theme) with the UX comments, It does not make sense to have huge files of the canvas theme if we are not using them and will be easier to maintain in the future.

What do you think @abonas ?

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -57,7 +57,7 @@
 		</div>
 		<!-- Header
 ============================================= -->
-		<header id="header" class="full-header transparent-header dark" data-sticky-class="not-dark">
+		<header id="header" class="full-header transparent-header" data-sticky-class="not-dark">
 
 			<div id="header-wrap">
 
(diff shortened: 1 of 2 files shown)


abonas:
I don't have any objections, it sounds like the right way to go. Would like to make sure with @theute because iirc he added the original theme/style.

aljesusg:
Hey @theute Can we merge this one?

theute:
Sure

PR #90, opened 2019-01-08
KIALI-2186 website - third level menu opens to the right and overlaps main menu

aljesusg:
This fix the problem but I want to check where the theme is adding this class because in the top level this class is added automatically @abonas 
peek 2019-01-08 15-37

### DIFF
--- a/themes/canvas/static/js/functions.js
+++ b/themes/canvas/static/js/functions.js
@@ -1053,16 +1053,21 @@ var SEMICOLON = SEMICOLON || {};
 		},
 
 		menuInvert: function() {
+			$('#primary-menu .mega-menu-content, #primary-menu ul ul ul').each( function( index, element ){
+				/*
+					Calculate the space on the right to check if the menu will be displayed outside window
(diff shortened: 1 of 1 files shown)


layouts/partials/header.html
```diff
@@ -95,7 +95,7 @@
 																<a href="{{ .URL }}">
 																	{{ .Name }}
 																</a>
-																<ul class="sub-menu">
+																<ul class="sub-menu menu-pos-invert">
 																	{{ range .Children }}
```
aljesusg: We can see here the change because is in the right,

```html
<ul class="sub-menu menu-pos-invert">
```
the added class is **menu-pos-invert**

abonas:
you are a magician :)
I wonder if we need all those checks in Kiali case, the menu is on right side of the website, and is opening to the left and there's always plenty of room.

aljesusg:
This is because I made the proposal of the new theme for Kiali, there are many methods and logic is that we do not need

PR #91, opened 2019-01-11
KIALI-2208 Adds an arrow to menus with submenus

josejulio:
edit: Updated screenshot with latest changes
image

Also did a small refactor to remove repeated code.

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -76,56 +76,7 @@
 			============================================= -->
 					<nav id="primary-menu" class="with-arrows clearfix">
 						<ul>
-							{{ range .Site.Menus.main }} {{ if .HasChildren }}
-							<li>
-								<a href="{{ .URL }}">
-									<div>{{ .Name }}</div>
-								</a>
-								<ul class="sub-menu">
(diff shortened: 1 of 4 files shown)


aljesusg:
What do you think about to flip the arrow from icon-angle-right to icon-angle-down when the submenu is collapse @josejulio @abonas ?

abonas:
I'm wondering if the arrow should be "down" or "left"? most of the menus are opening to the left and not down, only the main menu is opened "down"

josejulio:
This is how it looks like with a left arrow in the submenus.

It looks strange to me.

image

josejulio:
> What do you think about to flip the arrow from icon-angle-right to icon-angle-down when the submenu is collapse @josejulio @abonas ?

I think it looks good better if the arrow starts down and ends up to the left

peek 2019-01-14 11-50

josejulio:
I didn't push any change related to the previous screnshots, let me know what you think

josejulio:
Updated the code to reflect the previous gif

josejulio:
Are there more comments or changes that this PR needs?

PR #93, opened 2019-01-15
KIALI-2185 Allow VS to have empty weights in case actual weight sum is 100

xeviknal:
Documentation update after @amyroh improvement in https://github.com/kiali/kiali/pull/758

### DIFF
--- a/content/documentation/overview/index.adoc
+++ b/content/documentation/overview/index.adoc
@@ -192,13 +192,13 @@ This section lists all the validations that Kiali performs on all Istio configur
 
 |(Deprecated) Weight sum should be 100
 |error
-|When the sum of all the weight for a protocol doesnt sum 100
+|When the sum of all the weights for a protocol doesn't sum up to 100
 |https://github.com/kiali/kiali/tree/master/business/checkers/virtual_services/route_checker.go[source code, window="_blank"]
 |link:/files/validation_examples/109.yaml[109.yaml, window="_blank"]
 
 |(Deprecated) All routes should have weight
 |error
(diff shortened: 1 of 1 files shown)


content/documentation/overview/index.adoc
```diff
@@ -197,7 +197,7 @@ This section lists all the validations that Kiali performs on all Istio configur
 |link:/files/validation_examples/109.yaml[109.yaml, window="_blank"]
 
 |(Deprecated) All routes should have weight
```
abonas: I'm probably missing something here, but why is something under "deprecated" section fixed?

xeviknal: Deprecated validations are those validations that Galley is going to take care to warn users. But since not many people have Galley up and running, we decided to keep them listed.
Perhaps, we should explain that right at the bottom of the section, shall we?

abonas: why not many users have Galley up and running? is it disabled by default?

xeviknal: It is not disabled by default but in different installations they have problems to having properly running. Using both `oc cluster up` and `minishift start` have problems. Although, maistra team sorted it out on their installer.

abonas: is this a known/open issue in Istio repo? (shouldn't block this PR of course)

xeviknal: It has been many, I guess they are close though. Sooner or later we will have to revisit this decision about assuming galley is not working in some environments.
I did submit a PR to Istio to fix this: https://github.com/istio/istio/pull/11586

PR #98, opened 2019-02-01
KIALI-1548 - document uninstall in the getting started page

jmazzitelli:
https://issues.jboss.org/browse/KIALI-1548

### DIFF
--- a/content/gettingstarted/index.adoc
+++ b/content/gettingstarted/index.adoc
@@ -100,6 +100,24 @@ Once this is done, Kiali will be deployed and running in Kubernetes with an Ingr
 
 If you deployed in minikube, you can go directly to the Kiali UI at the URL `https://[minikube-ip]/kiali/console` where `[minikube-ip]` is what is reported when you run the command: `minikube ip`.
 
+== Uninstall
+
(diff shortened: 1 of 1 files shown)


content/gettingstarted/index.adoc
```diff
@@ -100,6 +100,24 @@ Once this is done, Kiali will be deployed and running in Kubernetes with an Ingr
...
 
+== Uninstall
+
+To uninstall Kiali is simple - just delete the objects that have the label `app=kiali` from your cluster.
+
+To uninstall from OpenShift:
+
+[source,bash]
+----
+oc delete all,secrets,sa,configmaps,deployments,ingresses,clusterroles,clusterrolebindings,virtualservices,destinationrules,customresourcedefinitions,templates --selector=app=kiali -n istio-system
```
lucasponce: @jotak minor question, with the customresourcedefinitions do we completely remove the runtimes info ?
I think yes, but I'd like to confirm just if here it needs an additional comment.

jotak: yes, at least this is what i've always experienced while redeploying locally: removing CRD also removes the associated resources.

PR #99, opened 2019-02-01
Update destination rule validations according current master state

xeviknal:
Updating validation documentation. DR were updated.

screenshot of kiali 1

### DIFF
--- a/content/documentation/overview/index.adoc
+++ b/content/documentation/overview/index.adoc
@@ -131,11 +131,17 @@ This section lists all the validations that Kiali performs on all Istio configur
 |https://github.com/kiali/kiali/tree/master/business/checkers/destination_rules/multi_match_checker.go[source code, window="_blank"]
 |link:/files/validation_examples/001.yaml[001.yaml, window="_blank"]
 
-|Host doesn't have a valid service
+|This host has no matching workloads
 |error
 |When one destination rule has a host that doesn't exist. This checks against any workload, service names as well as service entries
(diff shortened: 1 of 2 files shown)


abonas:
the explanation sounds the same in first and third columns, it is exactly the same words just said a bit in a different way. what is the added value of having both?

xeviknal:
First column is just the message that users receive. The thirds column is a broader (if needed) explanation of that message.

abonas:
> First column is just the message that users receive. The thirds column is a broader (if needed) explanation of that message. I think is useful to add extra messages to clarify users doubts. Probably we will have to add more verbosity and details to third column though.

yes, otherwise it's just paraphrasing and it looks a bit strange.

content/documentation/overview/index.adoc
```diff
@@ -131,11 +131,17 @@ This section lists all the validations that Kiali performs on all Istio configur
...
 
-|Host doesn't have a valid service
+|This host has no matching workloads
 |error
 |When one destination rule has a host that doesn't exist. This checks against any workload, service names as well as service entries
-|https://github.com/kiali/kiali/tree/master/business/checkers/destination_rules/no_host_checker.go[source code, window="_blank"]
+|https://github.com/kiali/kiali/blob/master/business/checkers/destinationrules/no_dest_checker.go#L28[source code, window="_blank"]
 |link:/files/validation_examples/002.yaml[002.yaml, window="_blank"]
+
+|This subset's labels are not found from any matching host
```
abonas: "not found from any" sounds a bit strange. maybe "on any ..." or "in any..." ?

xeviknal: Changed in kiali/kiali too.

PR #100, opened 2019-02-05
Update outdated demo video on homepage

xeviknal:
Homepage video is from version 0.3. We are at 0.14. I guess it would be great to replace it for a fresher one. I just found this vid from @pilhuhn (from mid october).
This is a wdyt PR, so wdyt?


mtho11:
I'm working on a more up to date one myself. But this is a great move forward.

PR #101, opened 2019-02-06
KIALI-2358 Update doc to include updated service account name

mwringe:
The service account name was updated a while ago to match what is being used in upstream Istio. But the docs for running with reduced permissions was not updated at that time.

### DIFF
--- a/content/gettingstarted/index.adoc
+++ b/content/gettingstarted/index.adoc
@@ -185,6 +185,6 @@ Then you will need to grant the `kiali` role in the namespace of your choosing:
 
 [source,bash]
 ----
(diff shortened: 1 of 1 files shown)


jmazzitelli:
Uhh.. I'm not sure this is true anymore. I think the service account name is "kiali" to match all the rest of Kiali's resource names. We should verify and make sure we still want kiali-service-account

mwringe:
@jmazzitelli We need to keep the service account name as 'kiali-service-account' since this matches what upstream Istio and Maistra is doing. If we have renamed this in the operator we probably need to change it back

jmazzitelli:
FYI: upstream is going to be using the operator soon (I have that on my todo list) and so should maistra (that's a JIRA on that). So in the future, we are going to be able to dictate the names of Kiali resources.

PR #106, opened 2019-03-04
KIALI-2457 Add new validation on DestinationRules about meshwide mTLS

xeviknal:
Adding documentation for https://github.com/kiali/kiali/pull/888


abonas:
I think that it'll be good to have an example of how to *fix* a validation problem, right now it's the examples of the validation issue itself and not its solution.

PR #109, opened 2019-03-07
KIALI-2122 Document runtimes monitoring

jotak:
- Add section in Features page
- Add full page documentation
- Increasing size of images in Features page
- Fix some CSS errors (reported by my IDE)

Features page:

features

The documentation page itself is quite long, can't show screenshot of everything, here's just the beginning:

doc


jotak:
cc @JStickler (you may want to have a look, too )

jotak:
By the way, I increased the size of the screenshots in "Features" page, I think they were really not easy to read (at least for me, and my screen isn't so small), please tell me if you think it's fine or not.

content/documentation/runtimes-monitoring/index.adoc
```diff
@@ -0,0 +1,228 @@
...
+== Pods Annotations
+
+Application pods must be annotated, both for the Prometheus scraper and to associate with Kiali dashboards. For instance, within a _Deployment_ definition:
+
+```yaml
+spec:
+  template:
+    metadata:
+      annotations:
+        prometheus.io/scrape: "true"
```
pilhuhn: Try putting  `#1` etc behind the lines and then below for the explanations you can reference
via `<1>` etc.
See https://asciidoctor.org/docs/user-manual/#callouts

content/documentation/runtimes-monitoring/index.adoc
```diff
@@ -0,0 +1,228 @@
...
+Additionally, you have to tell Kiali what is the URL of that Prometheus server: either by setting the environment variable `PROMETHEUS_CUSTOM_METRICS_URL` or Kiali _ConfigMap_:
+
+```yaml
+# ...
+external_services:
+  prometheus_custom_metrics_url: URL_TO_SERVER
+# ...
+```
+
+Make sure that Kiali pod can reach this URL.
```
pilhuhn: Can we describe how to do that ?

content/documentation/runtimes-monitoring/index.adoc
```diff
@@ -0,0 +1,228 @@
...
+kubectl apply -f mydashboard.yml
+```
+
+Or for OpenShift:
+
+```bash
+oc apply -f mydashboard.yml
+```
+
+To make them available system-wide, just create them in Kiali namespace (usually _istio-system_). Else, they will be available only for Apps or Workloads of the same namespace. In the case where the same dashboard name exists in a specific namespace and in Kiali namespace, the former takes precedence.
```
pilhuhn: Shouldn't at the end "Kiali ns" be "istio-system" ? Or if you make this 'location independent', then change istio-system above to Kial-ns

content/documentation/runtimes-monitoring/index.adoc
```diff
@@ -0,0 +1,228 @@
...
+** *aggregator*: defines how the time-series are aggregated when several of them are returned for a given metric and labels set. For example, if a Deployment creates a ReplicaSet of several Pods, you will have at least one time-series per Pod. Since Kiali shows the dashboards at the Workload (ReplicaSet) level or at the App level, they will have to be aggregated. This field can be used to fix the aggregator, with values such as _sum_ or _avg_ (full list available link:https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators[in Prometheus documentation]). However, you can omit it: in that case, the aggregator will default to _sum_ and can be changed from the dashboard UI.
+** *aggregations*: list of labels eligible for aggregations / groupings (they will be displayed in Kiali through a dropdown list)
```
jotak: while writing this doc, I feel the word "aggregations" is not really accurate here, I wonder if I should rename to something like "grouping", as it allows to group series by selected labels.
Renaming here would mean change the data model in Kiali.

pilhuhn:
On 7 Mar 2019, at 18:15, Joel Takvorian wrote:
> really accurate here, I wonder if I should rename to something like 
> "grouping", as it allows to group series by selected labels.
> Renaming here would mean change the data model in Kiali.

When grouping is more clear, then use it. Now it is still easy
to change it.

content/documentation/runtimes-monitoring/index.adoc
```diff
@@ -0,0 +1,228 @@
...
+        - label: "method"
+          displayName: "Method"
+    - name: "Server active connections"
+      unit: ""
+      spans: 6
+      metricName: "vertx_http_server_connections"
+      dataType: "raw"
+```
+
+The *name* field (from metadata) corresponds to what you will have to set in pods annotation `kiali.io/runtimes`.
```
JStickler: Maybe add a link on the word "kiali.io/runtimes" back up to where you described this in 2. Pods Annotations?   Also it might be helpful to have this example match the example in the Pods Annotations (right now one is Thorntail and the other is vertx).

content/documentation/runtimes-monitoring/index.adoc
```diff
@@ -0,0 +1,228 @@
...
+
+Spec fields definitions are:
+
+* *runtime*: name of the related runtime. It will be displayed on the corresponding Workload Details page. Can be omitted to not display anything.
```
JStickler: .. .on the corresponding Application or Workload Details page.  (Assume it would display on both?)

content/documentation/runtimes-monitoring/index.adoc
```diff
@@ -0,0 +1,228 @@
...
+* *title*: dashboard title, displayed as a tab in App or Workloads Details
```
JStickler: Application (spell it out)

content/features/50-runtimes-monitoring.adoc
```diff
@@ -0,0 +1,15 @@
+---
+title: "Runtimes Monitoring"
+date: 2019-03-06T13:55:38+02:00
+draft: false
+subTitle: "Monitor your application metrics within Kiali"
+image: features/runtimes_monitoring.png
+weight: 50
+---
+
+Kiali comes with some default dashboards for a couple of runtimes: Thorntail, Vert.x, Spring Boot and Node.js.
```
JStickler: Same comment as above about alphabetizing the list.   Node.js, Spring Boot, Thorntail, Vert.x.

content/documentation/runtimes-monitoring/index.adoc
jotak: I can do that, but the drawback is that the code is not copy-paste friendly anymore

content/documentation/runtimes-monitoring/index.adoc
jotak: I don't understand your remark. Technically speaking the namespace to install cluster-wide dashboard is Kiali/Istio's own namespace, but I just specified that this namespace is usually "istio-system", unless configured differently. So it seems correct to me to refer to Kiali namespace.

jotak:
@pilhuhn 
> When grouping is more clear, then use it. Now it is still easy to change it.

Ok, but I'm still not sure about the wording. Is "groupings" correct English?

If we don't change it right now, we can still do it later, let the spec settle down in case there is more feedback, and then we'll use a new version like v1beta1 (it's currently v1alpha1) and Kiali will accept both.

pilhuhn:
> I can do that, but the drawback is that the code is not copy-paste 
> friendly anymore

According to 
https://asciidoctor.org/docs/user-manual/#copy-and-paste-friendly-callouts 
it is.

jotak:
> According to https://asciidoctor.org/docs/user-manual/#copy-and-paste-friendly-callouts it is.

@pilhuhn just tried, it doesn't work :-(
Even with a comment prefix `# <1>` doesn't work. Something must be wrong with our renderer.

content/documentation/runtimes-monitoring/index.adoc
```diff
@@ -0,0 +1,228 @@
...
+
+* _prometheus.io/scrape_ tells Prometheus to fetch these metrics or not
+* _prometheus.io/port_ is the port under which metrics are exposed
+* _prometheus.io/scheme_ must be set to http for non-secure or https for secure connection
+* _prometheus.io/path_ is the endpoint path where metrics are exposed, default is /metrics
+* _kiali.io/runtimes_ is a coma-separated list of runtimes / dashboards that Kiali will look for
+
+== Default dashboards
+
+Since version 0.15, Kiali comes with a set of default dashboards for three different runtimes: Vert.x, Thorntail and Node.js. In Kiali 0.16, a fourth one is added: Spring Boot.
```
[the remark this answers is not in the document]
jotak: Ok - so I'll forget my subjective and biased ordering :-)

pilhuhn:
On 8 Mar 2019, at 10:10, Joel Takvorian wrote:

> @pilhuhn just tried, it doesn't work :-(
> Even with a comment prefix `# <1>` doesn't work. Something must be 
> wrong with our renderer.

That's a pity. Then leave as is.

content/documentation/runtimes-monitoring/index.adoc
jotak: Ok, I keep the vertx example as it is more complete (includes more different types of data)

content/documentation/runtimes-monitoring/index.adoc
jotak: actually not, the current Application page wasn't really suited for that, it may have to be redesigned to show it (I mean, the runtimes metrics tabs are present, but not the mention of the runtime name in description)

content/documentation/runtimes-monitoring/index.adoc
jotak: I don't want to open the pandora box :-) There can be so many different situations I guess.. prometheus in-cluster or not, behind a firewall or not, etc. I think at the moment the user has his own custom infrastructure, he is responsible for establishing the connections.

jotak:
@JStickler I think comments are addressed

content/documentation/runtimes-monitoring/index.adoc
jshaughn: minor: "when several of them are returned for a given metric and labels set" -> "when several are returned for a given metric and label set"
minor: "However, you can omit it: in that case, the" -> "However, if omitted the"

As for "aggregations", it's not perfect because an aggregation is typically a result.  But alternatives aren't obvious.  Prometheus uses "by", which is sort or short for "groupBy".   Can you select only one label from the list, or several?  If one then "groupBy" or "groupings" could work.  If you can select multiple then "groupBy", "groupByLabels" or "groupingLabels" could work.    I'm OK if you stick with "aggregations" although I can understand if you want to change to one of the "group" variants.

content/documentation/runtimes-monitoring/index.adoc
```diff
@@ -0,0 +1,244 @@
...
+kubectl apply -f mydashboard.yml
+```
+
+Or for OpenShift:
+
+```bash
+oc apply -f mydashboard.yml
+```
+
+To make them available cluster-wide, just create them in Kiali namespace (usually _istio-system_). Else, they will be available only for applications or workloads of the same namespace. In the case where the same dashboard name exists in a specific namespace and in Kiali namespace, the former takes precedence.
```
jshaughn: minor: "To make them" -> "To make the dashboard resources"

content/features/50-runtimes-monitoring.adoc
```diff
@@ -0,0 +1,15 @@
+---
+title: "Runtimes Monitoring"
+date: 2019-03-06T13:55:38+02:00
+draft: false
+subTitle: "Monitor your application metrics within Kiali"
+image: features/runtimes_monitoring.png
+weight: 50
+---
+
+Kiali comes with some default dashboards for a couple of runtimes: Go, Node.js, Spring Boot, Thorntail and Vert.x.
```
jshaughn: minor: "Kiali comes with some default dashboards for a couple of" -> "Kiali comes with default dashboards for several"

jotak:
@jshaughn thanks, done
I'm keeping the "aggregations" word for the time being, anyway I prefer to change it in the model first, before the doc.

content/documentation/runtimes-monitoring/index.adoc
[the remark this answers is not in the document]
JStickler: If there's a reason for the order, keep it!   As a reader, it just makes it easier to find the thing you're looking for in a long list if the list makes sense (matches the UI, alphabetical, or some other reason).  =)

content/documentation/runtimes-monitoring/index.adoc
JStickler: OK, wasn't sure if you could add to both pages or not.

PR #113, opened 2019-03-20
KIALI-1023 Mention supported versions

pilhuhn:


content/gettingstarted/index.adoc
```diff
@@ -16,7 +16,11 @@ toc::[]
 :icons: font
 :imagesdir: /images/gettingstarted/
 
-icon:bullhorn[size=2x]{nbsp} Kiali currently requires Istio version *1.0*.
+icon:bullhorn[size=2x]{nbsp} Kiali works with Istio versions *1.0.5+* and *1.1*.
+ + 
+Kiali will support the latest Istio version (1.1 at the moment) and the development
+is closely following the development of Istio.
+Older Istio versions like 1.0.x may continue to work, but will no longer be actively supported.
```
abonas: I see this got merged now, but there's something confusing here. on one hand it says on line 19 that it supports 1.0.5 and above and 1.1.
But on the other hand on line 23 it says that older versions like 1.0.x may continue to work but not supported.
But 1.0.5 is also 1.0.x

abonas: so which one is correct? @pilhuhn

jmazzitelli: > "Older Istio versions like 1.0.x may continue to work, but will no longer be actively supported."

This is a true statement. Even 1.0.2, 1.03, or 1.0.4 may continue to work (operative word is *MAY*) but is not actively supported. In other words, don't be surprised if you find some thing(s) broken when working with 1.0.x versions (ANY 1.0.x versions).

PR #116, opened 2019-04-09
KIALI-2683 change gettingstarted - use kiali operator now

jmazzitelli:


content/gettingstarted/index.adoc
```diff
@@ -22,123 +22,125 @@ Kiali will support the latest Istio version (1.1 at the moment) and the developm
...
 ----
 oc login -u system:admin
 ----
 
-You can now install Istio if needed. See the link:https://istio.io/docs/setup/[Istio Setup docs] for details.
-
-=== Install Kiali
-
-Now install the latest release of Kiali.
+=== Kubernetes Preparation
```
abonas: it looks a bit as if this is a followup/sequential step after openshift preparation and not instead of it. the numbering also supports that. it will be good to emphasize it's only one of them

content/gettingstarted/index.adoc
```diff
@@ -22,123 +22,125 @@ Kiali will support the latest Istio version (1.1 at the moment) and the developm
...
 
-Now install the latest release of Kiali.
+icon:lightbulb[size=1x]{nbsp} The Kiali Operator install script can be configured by setting environment variables when running it. For example, if you do not want it to install a Kiali CR (and thus avoid having the Kiali Operator install Kiali itself), set the OPERATOR_INSTALL_KIALI environment variable to "false" and then run the script. Read the comments at the top of the link:https://github.com/kiali/kiali/blob/master/operator/deploy/deploy-kiali-operator.sh[Kiali Operator install script] for details on what custom configuration settings are available to you.
```
abonas: this part is not very clear. if someone doesn't want Kiali, then they still install the Kiali operator and disable something? what's the purpose?

content/gettingstarted/index.adoc
```diff
@@ -22,123 +22,125 @@ Kiali will support the latest Istio version (1.1 at the moment) and the developm
...
+To install the Kiali Operator (and, by default, the Kiali CR), run the install script via this command:
 
 [source,bash]
 ----
-bash <(curl -L http://git.io/getLatestKialiKubernetes)
+bash <(curl -L https://git.io/getLatestKialiOperator)
 ----
 
-By default, the installation script will prompt you to enter a username and passphrase for the credentials that you want users to enter in order to log in successfully to Kiali. The installation script can be customized through environment variables - see link:http://git.io/getLatestKialiKubernetes[the comments at the top of the install script].
+You will be prompted for the authentication strategy you want to use. Choose "login" if you want to define a username and password that users will be required to access Kiali. Choose "anonymous" to allow full access to Kiali without requiring any credentials. Choose "openshift" to use the OpenShift OAuth login which controls access based on the individual's RBAC roles in OpenShift (this is only a feature available when deployed on OpenShift).
```
abonas: what happens on k8s if someone chooses openshift? or they won't have this option/it will output an error?

content/gettingstarted/index.adoc
```diff
@@ -22,123 +22,125 @@ Kiali will support the latest Istio version (1.1 at the moment) and the developm
...
+If on OpenShift, you can easily find the URL to access Kiali by going to the OpenShift console and navigating to the `istio-system` project's Overview page:
+
+[#img-openshift]
+.OpenShift console
+image::os-console.png[OpenShift console]
+{nbsp} +
+
+In this case, the Kiali URL is `http://kiali-istio-system.192.168.64.13.nip.io`. In your case this will be a different IP.
+
+You can also use the `oc` (or `kubectl` if on Kubernetes) command to determine the base URL:
```
abonas: detecting something similar on k8s?

content/gettingstarted/index.adoc
jmazzitelli: Yes, this is going to be confusing for anyone who isn't familiar with operators. I can't figure out a good way to word this. But I'll try again.

This is actually the reason why I added the "What is the Kiali Operator?" section - to try to introduce operators which would then hopefully make it more understandable.

What this is trying to say is - if you don't have a specific Kiali configuration that you want (maybe you just want to kick the tires with a demo and just quickly get something up and running? Maybe you want to test Kiali running with all of its settings set to their defaults?), then this script can install the operator AND kiali together for you. If, on the other hand, you have a specific configuration you want Kiali to have, then don't tell this script to install Kiali for you. Just install the operator and then you create the Kiali CR with the configuration you want and deploy that Kiali CR manually yourself.

In other words, this script will provide you with a base Kiali installation *only as a convienence* (i.e. so you don't have to create the Kiali CR yaml and deploy it yourself). But, if the base "all defaults" Kiali installation is configured in a way you don't want, then you can just tell this script to only install the operator and you will tell the operator later how you want kiali installed.

For example, you may want to install Kiali with a specific Jaeger URL, Grafana URL, web root, and you want to disable metrics and you want to enable view_only mode. Well, you want a very specifically configured Kiali - not this "base" Kiali that uses all defaults. In this case, tell this install script to only install the operator and skip the Kiali installation. Then you can write your own Kial CR yaml file (with all those settings you want to customize) and then you can manually "oc create -f my-kiali-cr.yaml -n kiali-operator" which will then tell the operator to install kiali with that special configuration.

Now I just have to figure out how to word all this properly :)

jmazzitelli: I added a little bit extra here.

content/gettingstarted/index.adoc
jmazzitelli: I added a blurb here to explain what happens.

content/gettingstarted/index.adoc
jmazzitelli: I added a sentence here to say go through the k8s Ingress. I don't have a running k8s instance, so I can't give exact commands/details.

content/gettingstarted/index.adoc
jmazzitelli: I tweeked this section to hopefully make things more clear

jmazzitelli:
The kiali operator is in master now and the old scripts are gone. We need to merge this and we need to make another release of Kiali.

content/gettingstarted/index.adoc
```diff
@@ -16,129 +16,132 @@ toc::[]
...
 is closely following the development of Istio.
 Older Istio versions like 1.0.x may continue to work, but will no longer be actively supported.
 
-== Getting started on OpenShift
+icon:lightbulb[size=1x]{nbsp}For several commands listed on this page, the OpenShift CLI command `oc` is used to interact with the cluster environment. If you are on Kubernetes, simply replace `oc` with `kubectl` unless otherwise noted.
 
-To deploy Kiali to your OpenShift cluster you can do the following.
+== Installing Kiali
 
 === Preparation
```
JStickler: Would "Prerequisites" be a better heading here?

content/gettingstarted/index.adoc
```diff
@@ -16,129 +16,132 @@ toc::[]
...
+If you are not installing on OpenShift but instead installing on Kubernetes, there are no additional preparation steps required unless you are installing on Minikube. When installing on Minikube, ensure that you enable the Ingress add-on by executing the following command:
 
 [source,bash]
 ----
-bash <(curl -L http://git.io/getLatestKiali)
+minikube addons enable ingress
 ----
 
-By default, the installation script will prompt you to enter a username and passphrase for the credentials that you want users to enter in order to log in successfully to Kiali. The installation script can be customized through environment variables - see link:http://git.io/getLatestKiali[the comments at the top of the install script].
+When Kiali is installed in Minikube, this Ingress add-on will enable you to go directly to the Kiali UI at the URL `https://[minikube-ip]/kiali/console` where `[minikube-ip]` is what is reported when you run the command: `minikube ip`.
```
JStickler: >will enable you to go directly to the Kiali UI 

...will let you access the Kiali console at the URL....

content/gettingstarted/index.adoc
```diff
@@ -16,129 +16,132 @@ toc::[]
...
 
-Once the above has completed and the Docker image has been pulled from Dockerhub, go to the OpenShift console, select the istio-system project and determine the base-URL of Kiali
+=== Install Istio
 
-[#img-openshift]
-.OpenShift console
-image::os-console.png[OpenShift console]
-{nbsp} +
+Make sure your cluster environment has Istio installed. Because you are following these instructions to install the latest Kiali, you should not install the Kiali that comes with Istio (i.e. do not use the Helm setting `kiali.enabled=true` when installing Istio).
```
JStickler: Because you are following these instructions to install the latest version of Kiali, you should not install the Kiali version that is bundled with Istio (that is, do not use the Helm setting `kiali.enabled=true` when installing Istio).

content/gettingstarted/index.adoc
```diff
@@ -16,129 +16,132 @@ toc::[]
...
+==== What is the Kiali Operator?
 
-To deploy Kiali to your Kubernetes cluster you can do the following.
+First, a quick introduction of what the Kiali Operator actually is.
 
-=== Preparation
+The Kiali Operator is an implementation of a link:https://coreos.com/operators/[Kubernetes Operator]. It provides a method of packaging, deploying, and managing Kiali.
 
-Preparation is only needed if you are installing in Minikube. When installing in Minikube, ensure that you enable the Ingress add-on by executing the following command:
+The Kiali Operator runs inside the `kiali-operator` namespace and listens for the creation of a Kiali CR (custom resource) object. When it sees one created, it will install Kiali using the settings it finds in the CR object. To see all the different settings you can configure in a Kiali CR, see link:https://github.com/kiali/kiali/blob/master/operator/deploy/kiali/kiali_cr.yaml[this fully documented example Kiali CR yaml file]. A minimal Kiali CR that falls back to all defaults looks like this:
```
JStickler: >A minimal Kiali CR that falls back to all defaults looks like this:
A minimal Kiali CR that uses all the defaults looks like this:


(Also yay fully documented example!  \o/)

content/gettingstarted/index.adoc
```diff
@@ -16,129 +16,132 @@ toc::[]
...
+kind: Kiali
+metadata:
+  name: kiali
 ----
 
-You can now install Istio if needed. See the link:https://istio.io/docs/setup/[Istio Setup docs] for details.
+==== Installing the Kiali Operator
 
-=== Install Kiali
+In the instructions below, you will be shown how to install the Kiali Operator using the link:https://github.com/kiali/kiali/blob/master/operator/deploy/deploy-kiali-operator.sh[Kiali Operator install script]. That install script has the option of also immediately installing a simple Kiali CR for you so you do not have to, thus making it easy to get started with Kiali by running a single command (this is, in fact, the install script's default behavior). If, however, you wish to install Kiali specifically configured with custom settings of your choosing, you can tell the script to skip the installation of a Kiali CR and just simply install the Operator. In this case, you would be responsible for writing the Kiali CR yaml file and creating it in your cluster, thus allowing you to tell the Operator to install Kiali with the settings you want.
```
JStickler: >In the instructions below, you will be shown how to install the Kiali Operator using the 

In the instructions below, you will install the Kiali Operator using the ...

content/gettingstarted/index.adoc
```diff
@@ -16,129 +16,132 @@ toc::[]
...
+To install the Kiali Operator (and, by default, the Kiali CR), run the install script via this command:
 
 [source,bash]
 ----
-bash <(curl -L http://git.io/getLatestKialiKubernetes)
+bash <(curl -L https://git.io/getLatestKialiOperator)
 ----
 
-By default, the installation script will prompt you to enter a username and passphrase for the credentials that you want users to enter in order to log in successfully to Kiali. The installation script can be customized through environment variables - see link:http://git.io/getLatestKialiKubernetes[the comments at the top of the install script].
+You will be prompted for the authentication strategy you want to use. Choose "login" if you want to define a username and password that users will be required to access Kiali. Choose "anonymous" to allow full access to Kiali without requiring any credentials. Choose "openshift" to use the OpenShift OAuth login which controls access based on the individual's RBAC roles in OpenShift (this is only a feature available when deployed on OpenShift - the Operator will fail to install Kiali successfully and will log an appropriate error message if you choose this when installing on Kubernetes).
```
JStickler: Breaking this out into a bulleted list might make it easier to see how many options there are here.  What do you think?

* Choose "login" if you want to define a username and password that users will be required to access Kiali. 
* Choose "anonymous" to allow full access to Kiali without requiring any credentials. 
* Choose "openshift" to use the OpenShift OAuth login which controls access based on the individual's RBAC roles in OpenShift (this is only a feature available when deployed on OpenShift - the Operator will fail to install Kiali successfully and will log an appropriate error message if you choose this when installing on Kubernetes).

content/gettingstarted/index.adoc
```diff
@@ -16,129 +16,132 @@ toc::[]
...
 
-If you deployed in minikube, you can go directly to the Kiali UI at the URL `https://[minikube-ip]/kiali/console` where `[minikube-ip]` is what is reported when you run the command: `minikube ip`.
+icon:bullhorn[size=1x]{nbsp} If you told the install script to not create a Kiali CR (and thus not have Kiali installed yet) via the OPERATOR_INSTALL_KIALI=false environment variable, you are responsible for creating this secret if you wish to install Kiali with the authentication strategy of "login". A secret is not required if your authentication strategy is not "login". The following command is a simple way to create a secret for Kiali whose username is "admin" and passphrase is "admin":
+[source,bash]
+----
+oc create secret generic kiali -n istio-system --from-literal "username=admin" --from-literal "passphrase=admin"
+----
 
-== Uninstall
+Once the Kiali Operator has installed Kiali and the Kiali pod has successfully started, you can begin to access Kiali.
```
JStickler: >Once the Kiali Operator has installed Kiali and the Kiali pod has successfully started, you can begin to access Kiali.

Once the Kiali Operator has installed Kiali and the Kiali pod has successfully started, you can access the Kiali Console.

content/gettingstarted/index.adoc
```diff
@@ -16,129 +16,132 @@ toc::[]
...
+If on Kubernetes, access Kiali over the Ingress URL.
+
+== Uninstall
+
+To uninstall Kiali is simple - just delete the Kiali CR. This will trigger the Kiali Operator to uninstall Kiali:
 
 [source,bash]
 ----
-kubectl delete all,secrets,sa,configmaps,deployments,ingresses,clusterroles,clusterrolebindings,virtualservices,destinationrules,customresourcedefinitions --selector=app=kiali -n istio-system
+oc delete kiali kiali -n kiali-operator
```
JStickler: Should kiali be listed twice?

content/gettingstarted/index.adoc
jmazzitelli: "Prerequisites" makes it sound like I need some things already installed or some dependencies deployed. This really is how to prepare your OpenShift or Kubernetes environment in order to be able to run Kiali

content/gettingstarted/index.adoc
jmazzitelli: done (when I say done, I mean, its on my local copy - I'll commit all these changes when I go through all these comments)

content/gettingstarted/index.adoc
jmazzitelli: done (you should use the "insert suggestion" feature in github - its hard to figure out what you are changing and what is the same)

content/gettingstarted/index.adoc
jmazzitelli: done

content/gettingstarted/index.adoc
jmazzitelli: done

content/gettingstarted/index.adoc
jmazzitelli: done

jmazzitelli: blech.. at least on my local hugo server, the font size shrink when I use bulleted list. Oh, well. I'll leave it like that - I like the bulleted list even if the font is smaller.

content/gettingstarted/index.adoc
jmazzitelli: yes... the command is "oc delete <the type of resource> <the name of the instance to be deleted> ..." in this case the resource type and the resource name are the same -- "kiali"

content/gettingstarted/index.adoc
jmazzitelli: Rather than "Console" I will say "UI" because everywhere else in this page calls it the "Kiali UI"

content/gettingstarted/index.adoc
JStickler: Still things you need before you can run Kiali, but it's fine either way.  Not a hill I want to die on.  ;-)

content/gettingstarted/index.adoc
JStickler: I don't know about that feature.  *Googles*  "Available on Developer, Team, and Business Cloud plans"  Do we have one of those plans?

content/gettingstarted/index.adoc
JStickler: Ah, I was wondering if it was a typo or something like that, two resources with the same name.  Thanks!

PR #123, opened 2019-05-15
kiali-2861 Add new overview video

mtho11:
This just places an updated Kiali introductory video.

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -69,7 +69,7 @@
               </div>
               <div class="col_half topmargin-sm col_last" style="vertical-align: center">
                 <div style="position: relative; padding-bottom: 56.25%; height: 0; overflow: hidden;">
-                  <iframe src="//www.youtube.com/embed/6NWPi8lm2Y0" style="position: absolute; top: 0; left: 0; width: 100%; height: 100%; border:0;" allowfullscreen title="Kiali Overview"></iframe>
+                  <iframe src="https://player.vimeo.com/video/335512121" style="position: absolute; top: 0; left: 0; width: 100%; height: 100%; border:0;" allowfullscreen title="Kiali Overview"></iframe>
                 </div>
               </div>
             </div>


mtho11:
@abonas I can just as easily publish to youtube if that is the need. How do team members publish to Kiali Channel?

I used Vimeo instead of Youtube because the quality of video is better on vimeo than youtube; At least that has been the case before when text on screencasts was not as sharp. I can even publish 4K video.

Also with Vimeo I can update the videos in place, without changing the url. So updates that I make can be transparent to the website.

abonas:
@mtho11 @theute afaik has credentials to youtube. I don't really mind in which one the content is, but I'm concerned right now it's split. If more people are in favor of vimeo, perhaps we can migrate over there, but I'll let @theute speak

mtho11:
Just an FYI: my software publishes directly to a provider/account so it can best maximize the format/size/settings for that provider. I'm not just producing a webm file or whatever. Although I could and just lose some of the software setting smarts.

mtho11:
@theute @abonas IMO we should not mix the tutorial videos with the sprint demos. Most non-RH people may want to watch some quick tutorial videos to come up to speed but don't care about what we did on our sprints a few weeks ago.
I just think this needs some further thought as the tutorial videos should be in the same spot so someone can peruse those with the clutter of sprint demos.

josejulio:
Youtube allows you to group related videos (e.g. sprint demos, tutorial, etc). We could use that to avoid mixing them together and making it easy to find what you are looking for.

theute:
Yes, let's please have them all on one place...
We are slowly growing the number of followers on YouTube and it just makes sense to have all our published videos in one place.
We have a playlist for Sprints, we can create one for tutorials.

I invited you as "manager" of the channel so that you can add videos there.

(Vimeo definitely has nice(r) features, but the Basic plan is quite limited, YouTube does the job even though you can't indeed reupload a video without changing the URL)

mtho11:
@theute ok moving the videos over to youtube.

PR #127, opened 2019-05-24
KIALI-2189 remove installation section in favor of a consolidated getting-started

jshaughn:
Move a small amount of information into Getting Started from Installation and remove Installation section.


config.toml
```diff
@@ -33,35 +33,11 @@ theme = "canvas"
       weight = 2
       parent="documentation"
   [[menu.main]]
-      identifier = "doc_installation"
-      name = "Installation"
-      url = "/documentation/installation"
+      identifier = "doc_gettingstarted"
+      name = "Getting Started"
+      url = "/documentation/getting-started"
```
jmazzitelli: don't change this - we have "https://kiali.io/gettingstarted" URL all over external - tweets, upstream Istio docs, etc.

jshaughn: OK, that's a bit of a pain the way things are structured, but maybe I'll see what I can do to preserve that link.

content/documentation/getting-started/index.adoc
```diff
@@ -52,8 +53,11 @@ oc login -u system:admin
 
 === Install Istio
 
-Make sure your cluster environment has Istio installed.
-For details on how to install Istio, see the link:https://istio.io/docs/setup/[Istio Setup docs].
+Make sure your cluster environment has Istio installed. For details on how to install Istio, see the link:https://istio.io/docs/setup/[Istio Setup docs].  If you are running on OpenShift and prefer Maistra, see link:https://maistra.io/docs/getting_started/install/[Maistra Setup docs].
+
+icon:lightbulb[size=1x]{nbsp}Both Istio and Maistra include Kiali distributions.
+
+If the packaged version of Kiali is sufficient then choose the options to install Kiali and then continue to link:#_open_the_ui[Open the UI].  If you prefer to install the latest Kiali then do not choose the options to install Kiali, or delete it as necessary, and continue with the steps below.
```
jmazzitelli: This isn't clear you are talking about the packaged version **that comes with Istio or Maistra**. Say something like:

```
If the Kiali version that comes packaged with Istio or Maistra is sufficient for you, then you
may install Kiali using the Istio or Maistra installation procedures and then continue to
link:#_open_the_ui[Open the UI]. If you prefer to install the latest Kiali then do not have
Istio or Maistra install Kiali and continue with the steps below. Note that the Kiali installation
script documented below will check to see if Kiali is already installed and, if it is, will ask
to delete it.
```

config.toml
jshaughn: I found the way to alias previous urls so this should be fine.

content/documentation/getting-started/index.adoc
jshaughn: I updated the text

jmazzitelli:
LGTM assuming that alias works - it doesn't with the netlify test deployment, I'll assume this works when it gets to the real kiali.io site.

jshaughn:
>  it doesn't with the netlify test deployment, I'll assume this works when it gets to the real kiali.io site.

It worked with "make serve" but it's not working on the main site.  I'll try and fix.

PR #132, opened 2019-05-31
KIALI-2925 kiali.io: remove API from top menu

jshaughn:
Multiple people indicated that although the API is not for public clients,
having it published has helped some contributors. So, instead of removing it
from the website, this PR just moves the API option under Documentation.
- also, remove some obsolete layout files for Features


jshaughn:
@lucasponce We could rename the menu option from API to "Developer API" or something like that.  Would that make it any better for you?

lucasponce:
An adjective like "Developer API" will help, but also I guess we need some kind of note on the API.

Something to say that this is a internal API which is intended to be opaque for Kiali users and it can be changed between versions without guaranteed backward compatibility.

For example, k8s has similar comments in some fields, so I guess that should be also added in the project. 

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md

josejulio:
> just moves the API option under Documentation.

The link in the menu is broken: https://deploy-preview-132--kiali-io-draft.netlify.com/documentation/api

The link is also missing in the documentation index page:
https://deploy-preview-132--kiali-io-draft.netlify.com/documentation/

jshaughn:
@lucasponce Changed to "Developer API" but +1 to also changing the content of that page in another PR.

jshaughn:
@josejulio It is because the API doco is not generated in netlify.  It should be good on the real page. When it gets synced I'll check.

PR #148, opened 2019-07-04
KIALI-3087 Clarify how to fix known troubles in Jaeger

xeviknal:
Add two new entries to FAQ for Tracing.
1. Using https route with untrusted certificate.
2. User doesn't have access to jaeger query endpoint. Integration disabled.


josejulio:
See https://www.kiali.io/faq/ for the current FAQ

xeviknal:
@aljesusg @josejulio I moved the DT troubleshooting to FAQs.

PR #154, opened 2019-07-10
KIALI-3115 Service validation example and description improvements

burmanm:
Updates the documentation to match updates from the PR kiali/kiali#1230

### DIFF
--- a/data/files/validation_examples/703.yaml
+++ b/data/files/validation_examples/703.yaml
@@ -1,0 +1,41 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: ratings-java-svc
+  namespace: ratings-java
+  labels:
+    app: ratings
+    service: ratings-svc
+spec:
+  ports:
+  - port: 9080
+    targetPort: 8080
+    name: http
+  selector:
+    app: ratings-java
+    version: v1
+---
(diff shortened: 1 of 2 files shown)


data/files/validation_examples/703.yaml
```diff
@@ -0,0 +1,41 @@
...
+metadata:
+  name: ratings-java-svc
+  namespace: ratings-java
+  labels:
+    app: ratings
+    service: ratings-svc
+spec:
+  ports:
+  - port: 9080
+    targetPort: 8080
```
xeviknal: Is this example creating a warning/error? I understand that this example is valid, right? Shouldn't we modify this to make it fail? Afaik, all the other examples creates validations in order to make the user see where the problem is.
Perhaps understanding wrong the validation.

data/files/validation_examples/703.yaml
```diff
@@ -0,0 +1,41 @@
...
+    name: http
+  selector:
+    app: ratings-java
+    version: v1
+---
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: ratings-java
+  namespace: ratings-java
```
xeviknal: What do you think about removing the namespace? Or using bookinfo perhaps?

PR #155, opened 2019-07-15
Add documentation for label selector

johscheuer:
This PR add documentation for the `label_selector` feature discussed here: https://github.com/kiali/kiali/issues/1214

Fixes: https://github.com/kiali/kiali/issues/1214

### DIFF
--- a/content/documentation/getting-started/index.adoc
+++ b/content/documentation/getting-started/index.adoc
@@ -381,7 +381,7 @@ namespaces:
   label_selector: kiali-enabled=true
 ----
 
-For further information how the `label_selector` interacts with `deployment.accessible_namespaces` read the https://github.com/kiali/kiali/blob/v1.1.0/operator/deploy/kiali/kiali_cr.yaml#L44-L53[technical documentation].
+For further information on how the `label_selector` interacts with `deployment.accessible_namespaces` read the https://github.com/kiali/kiali/blob/master/operator/deploy/kiali/kiali_cr.yaml[technical documentation].
 
 In order to label a namespace you can use following command, for more information see the :link:https://kubernetes.io/docs/concepts/overview/working-with-objects/labels[official documentation]
 
--- a/content/documentation/getting-started/index.adoc
+++ b/content/documentation/getting-started/index.adoc
@@ -369,7 +369,11 @@ namespaces:
 
 ==== Namespace Selectors
 
-Kiali supports an optional label selector for namespaces which is used to fetch a subset of the available namespaces. The label selector is defined under the namespaces definition. The example below selects all namespaces that have a label `kiali-enabled: true`.
+Kiali supports an optional label selector for namespaces which is used to fetch a subset of the available namespaces.
+
+The label selector is defined under the namespaces definition.
+
+The example below selects all namespaces that have a label `kiali-enabled: true`:
 
 [source,yaml]
 ----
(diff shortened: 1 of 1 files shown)


content/documentation/getting-started/index.adoc
```diff
@@ -367,6 +367,17 @@ namespaces:
     - mycorp_test.*
 ----
 
+==== Namespace Selectors
+
+Kiali supports an optional label selector for namespaces which is used to fetch a subset of the available namespaces. The label selector is defined under the namespaces definition. The example below selects all namespaces that have a label `kiali-enabled: true`.
```
pilhuhn: ( While not the general style of the file, it is a good idea to start a new sentence on a new line  -- no need to change it now )

Would it make sense to show the command to label a namespace with that selector for people that are less familiar? Or link to the kube documentation?

pilhuhn:
Thank you for writing this up!

content/documentation/getting-started/index.adoc
```diff
@@ -367,6 +367,17 @@ namespaces:
...
+
+[source,yaml]
+----
+namespaces:
+  label_selector: kiali-enabled=true
+----
+
```
jmazzitelli: I would want to talk about the reason why we originally added this label_selector - and that is to support soft-multitenancy. Particularly, we need to tell people the value of that label selector should be unique across multiple meshes. My suggested addition is below:

```suggestion

Note that when using multiple service meshes (i.e. multiple control planes) in the same cluster, you will want to set the label selector's value to a value unique to each mesh. This is so each mesh's Kiali instance will only select those namespaces within that mesh. 
For an example of using Kiali in this kind of soft multi-tenancy mode, see the [Maistra](https://github.com/Maistra/istio-operator) project.
This is the reason why this `label_selector` will be defined by default to the value of `kiali.io/member-of: <istio_namespace>` if the `deployment.accessible_namespaces` is set to something other than the "all namespaces" value `['**']`. This allows you to have multiple control planes in the same cluster and have each control plane contain its own Kiali instance.

```

jmazzitelli:
FYI: @jshaughn will be unable to review for a couple weeks.

content/documentation/getting-started/index.adoc
johscheuer: I added it

content/documentation/getting-started/index.adoc
johscheuer: Updated the docs

content/documentation/getting-started/index.adoc
```diff
@@ -367,6 +367,38 @@ namespaces:
...
+The example below selects all namespaces that have a label `kiali-enabled: true`:
+
+[source,yaml]
+----
+namespaces:
+  label_selector: kiali-enabled=true
+----
+
+For further information how the `label_selector` interacts with `deployment.accessible_namespaces` read the https://github.com/kiali/kiali/blob/v1.1.0/operator/deploy/kiali/kiali_cr.yaml#L44-L53[technical documentation].
```
jshaughn: minor:   "information how" => " information on how"

Also, I'd avoid versioning and line numbers in the link.  At the moment we don't have a versioning solution in the docs (we need that).  And line number can be brittle.  How about:

read the 'label_selector' documentation in https://github.com/kiali/kiali/blob/master/operator/deploy/kiali/kiali_cr.yaml[kiali_cr.yaml].

israel-hdez:
I'm merging.
Thanks @johscheuer for your contribution.

PR #162, opened 2019-08-14
FAQ entry: How do I access Kiali?

israel-hdez:
Moving the instructions about accessing Kiali to a FAQ, and providing
more concrete instructions.

Also, relocating the "Open the UI" subsection of the Getting Started
guide to place it under the "Install Kiali Latest", instead of the
"Uninstall".


israel-hdez:
@jmazzitelli This is the FAQ entry you asked some days ago.
I haven't tested the OpenShift commands, becuase I don't have a cluster right now. Perhaps you could try?

israel-hdez:
@jmazzitelli What I understand from these section of the k8s docs (bullet 4): https://kubernetes.io/docs/tasks/access-application-cluster/ingress-minikube/#create-an-ingress-resource is that in k8s you could get the IP assigned to the ingress rule like this:

`kubectl get ingress kiali -n istio-system -o jsonpath='{.status.loadBalancer.ingress[0].ip}'`

But as pointed in that same part of those docs, in minikube this returns a private IP that is unreachable from the host. Not sure if that would work in any of the clouds out there...

israel-hdez:
Merging. I already got the :heavy_check_mark: .

PR #167, opened 2019-09-19
Graph FAQ entry supporting edge summary Hosts tab

jshaughn:
https://github.com/kiali/kiali/issues/1405

### DIFF
--- a/content/faq/graph/no-hosts.adoc
+++ b/content/faq/graph/no-hosts.adoc
@@ -1,0 +1,42 @@
+---
+title: Why does my edge summary "Hosts" tab say "no hosts configured"?
+weight: 80
+---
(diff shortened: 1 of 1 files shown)


josejulio:
I'm not sure, but given that this could be possible implemented later by Istio, maybe we should add a line that this is experimental and subject to changes. 
I'm not sure because anyway p8s won't store the metrics for too long and if Istio provides a out-of-the-box way, we just use it.

jshaughn:
Starting with Istio 1.3.1 it is not necessary to configure request.host as it is set as destination_service, which is a a standard telemetry field.  This is true only for egress traffic, but that is the primary use case.  So, this FAQ PR can be closed.

Issue #164, opened 2019-08-29
Prometheus setting for custom metrics is outdated


pavolloffay:
Is `prometheus_custom_metrics_url` it outdated?  https://www.kiali.io/documentation/runtimes-monitoring/

In my kiali configmap I have:

```
      prometheus:
        auth:
          ca_file: /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
          insecure_skip_verify: false
          password: 1X/CET+lKOm0t6R5UfcYdrRMJ1gc6oCbU8DJg/i4IxCHWJ4YrZAW1kRsMvUXGCEh8fYFvjo3QLdzcarJmdqjFtr7IktKRvCnR2hcxgXncibsiKoWcDukilb9qQyw7IlXAc8PWAhy5tRvTN2/tLKAcXFSXxAkifMnEg8QOMFQ504b1Qhi3OaBNTmqBtjYRKQNhetcwTX0l+SKS7zsKISdLig79b4l3lD6zzp9BZeTdAELi7NtceR8nzbGJmRov8LTNwhzr3un20GiUMv4tII5oz9aVKYcqmDWJdY5uPJJf0zBQuZKYI7WhLa1VkVelJ6oU7KQoHxNSYYMFGyouzB9
          token: ''
          type: basic
          use_kiali_token: false
          username: internal
        custom_metrics_url: https://prometheus.istio-system.svc:9090
        url: https://prometheus.istio-system.svc:9090
```

xeviknal:
I guess this issue can be closed per #169 :tada:

Issue #168, opened 2019-09-25
Fail to run site locally


jotak:
I haven't investigated deeply but our commands documented in readme don't seem to work anymore.

- `make install` is (almost) OK, I just need to `sudo`
- `make serve` fails with:
  - many errors (maybe not blocking) in `documentation/validations/index.adoc` `WARNING: include file is outside of jail; recovering automatically`
  - Also:

```
Error: Error building site: failed to render pages: render of "taxonomyTerm" failed: "/work/kiali.io/themes/kiali/layouts/_default/list.html.html:14:25": execute of template failed: template: _default/list.html.html:14:25: executing "content" at <where (where $pagesOfSection "Parent" "!=" nil) "Parent.File.Path" .File.Path>: error calling where: runtime error: invalid memory address or nil pointer dereference
make: *** [Makefile:14: serve] Error 255
```

- `make hugo-serve` also fails for me:

```
unknown flag: --mount
See 'docker run --help'.
make: *** [Makefile:33: hugo-serve] Error 125
```

I'm on Fedora 30, running docker 1.13.1, hugo v0.55.6

abonas:
@jotak is this still an issue or can be closed?

jotak:
Looks like it's fixed (fedora 31, hugo 0.59.1)
Warnings are still there, but the site is served & rendered.

PR #169, opened 2019-09-25
Update documentation on runtimes monitoring

jotak:
- New fields in CR: chartType, min/max, externalLinks
- Fix outdated prometheus config (#164)

Closes https://github.com/kiali/kiali/issues/1727


jotak:
Note that I couldn't deploy & test locally due to https://github.com/kiali/kiali.io/issues/168

PR #171, opened 2019-10-02
Adds a search input

josejulio:
- Redirects to google with site:www.kiali.io

Preview:  https://deploy-preview-171--kiali-io-ci.netlify.com/

Peek 2019-10-02 11-49
Peek 2019-10-02 11-45

But the current code doesn't work very well from width 1050 to 1245.

Peek 2019-10-02 11-48

Fixes kiali/kiali#1611 and replaces #157 
Also Closes kiali/kiali-design#142

### DIFF
--- a/themes/kiali/layouts/_default/baseof.html
+++ b/themes/kiali/layouts/_default/baseof.html
@@ -80,6 +80,22 @@
           }	
         })	
       });
+
+      // Search input
+      const searchForm = $('body > header div.search-container form');
+      searchForm.submit(function(event) {
+        event.preventDefault();
+        const searchInput = $(event.target).children('input[type=text]');
+        const href = $(event.target).attr('action');
+        const query = `site:www.kiali.io ${searchInput.val()}`;
+        window.open(`${href}?${searchInput.attr('name')}=${encodeURIComponent(query)}`, '_self');
+      });
+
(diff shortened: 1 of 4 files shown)


josejulio:
I did a small change that I forgot to address a comment:

> My thinking is that the black search icon and text should be the treatment when you have entered information in (whether you are in focus or not), and then the blue rectangle all around is what would indicate focus



Peek 2019-10-02 12-05

josejulio:
> But the current code doesn't work very well from width 1050 to 1245.

@serenamarie125 or @maryshak1996 could you take a look and suggest how to proceed?
I think that our options are removing the FAQ from the top menu and probably shrinking the font a little bit or jump earlier to the "mobile" top menu.

israel-hdez:
Special request. Since you are here, can you fix this style?: https://github.com/kiali/kiali.io/blob/master/themes/kiali/static/css/kiali-home.css#L247-L251

min-width should be 1500px.

josejulio:
> Special request. Since you are here, can you fix this style?: https://github.com/kiali/kiali.io/blob/master/themes/kiali/static/css/kiali-home.css#L247-L251

sure, no problem

josejulio:
@israel-hdez please re-review.
I added some changes suggested by UX.

PR #173, opened 2019-10-11
Experimenting with a google custom search engine

josejulio:
Google allows you to create a custom search engine https://cse.google.es/cse/

That way, you have more options to customize the search page, for example, I added the Kiali logo out there.
The page ends up being more simple, i'm unsure if we should use it or keep the current one.

Screenshot_2019-10-11_16-47-38

### DIFF
--- a/themes/kiali/layouts/_default/baseof.html
+++ b/themes/kiali/layouts/_default/baseof.html
@@ -82,14 +82,7 @@
       });
 
       // Search input
-      const searchForm = $('body > header div.search-container form');
-      searchForm.submit(function(event) {
-        event.preventDefault();
-        const searchInput = $(event.target).children('input[type=text]');
-        const href = $(event.target).attr('action');
-        const query = `site:www.kiali.io ${searchInput.val()}`;
-        window.open(`${href}?${searchInput.attr('name')}=${encodeURIComponent(query)}`, '_self');
-      });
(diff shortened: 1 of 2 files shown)


jshaughn:
I like how the default page includes images but this custom page makes it much more clear that you are on a site-specific search.  Can the custom page include images?

josejulio:
> Can the custom page include images?

It does, but in a different way than google.com.

Google shows by default a mixed content All ("Todos" in the gif)
Peek 2019-10-14 09-31

While the custom search engine shows them separated, I didn't see any option to have this mixed view.
Peek 2019-10-14 09-29

josejulio:
The interesting things I found are that they provide an API so you can include the results in your page and they provide options to change fonts and colors of the text.

israel-hdez:
I like better this approach, because people doesn't need to worry about the "site:kiali.io" keyword. And you can also click the Kiali logo to go back to the main site :)

jshaughn:
@josejulio @israel-hdez , so what is the consensus?  Shall we merge this change?

josejulio:
I'm +1 on merging this :)
The search page is more personalized and less hacky (you don't get to see the site:kiali.io)

israel-hdez:
I vote for merging.

josejulio:
Lets merge then, we can revisit this later if we need to customize any color or revert back to default google.

PR #187, opened 2019-12-11
Adding host format issues to a new FAQ

xeviknal:


content/documentation/validations/index.adoc
```diff
@@ -87,6 +87,9 @@ This validation warning could be a result of duplicate definition of the same su
 
 Istio silently ignores the duplicate subsets and merge these destination rules without letting the user know. Only the first seen rule (by Istio) per subset is used and information from multiple definitions is not merged. While the routing might work correctly, this is most likely a configuration error. It may lead to a undesired behavior if one of the offending rules is removed or modified and that is probably not the intention of the deployer of this service. Also, if the two offending destination rules have different policies for traffic routing the wrong one might be used.
 
+icon:exclamation-triangle[] Istio highly recommend to always use FQDN for hosts on DestinationRules. However it allows to use other formarts like short names ('details' or 'details.bookinfo').
```
JStickler: Istio highly recommends that you always use fully qualified domain names (FQDN) for hosts in DestinationRules.  However, Istio does allow you to use other formats like short names ('details' or 'details.bookinfo').

content/documentation/validations/index.adoc
```diff
@@ -87,6 +87,9 @@ This validation warning could be a result of duplicate definition of the same su
...
+Kiali only supports FQDN and simple service name as host formats: 'details.bookinfo.svc.cluster.local' or 'details' instead. Validations using the 'details.bookinfo' format might not be accurate.
```
JStickler: Kiali only supports FQDN and simple service names as host formats: for example, 'details.bookinfo.svc.cluster.local' or 'details'. 

(Should 'simple service names' be plural?)

"Validations using the 'details.bookinfo' format might not be accurate."  This implies that "details" is preferred over "details.bookinfo", is that what you intended?

content/faq/validations/host-format.adoc
```diff
@@ -0,0 +1,20 @@
+---
+title: Which formats does Kiali support for specifying hosts?
+weight: 70
+---
+:icons: font
+
+Istio highly recommend to always use FQDN for hosts on DestinationRules. However it allows to use other formarts like short names (_details_ or _details.bookinfo_).
```
JStickler: Same comments as above.

content/faq/validations/host-format.adoc
```diff
@@ -0,0 +1,20 @@
...
+
+In the following example it should show the validation of "More than one Destination Rule for the same host subset combination". Because of the usage of the short name _reviews.bookinfo_ Kiali won't show the warning message on both destination rules.
+
+[source, yaml]
+----
+include::/data/files/validation_examples/001-format-unsupported.yaml[]
+----
+
+See the recomendation Istio gives regarding https://istio.io/docs/reference/config/networking/destination-rule/#DestinationRule[host format^]:
+_"To avoid potential misconfigurations, it is recommended to always use fully qualified domain names over short names."_
```
JStickler: "To avoid potential misconfigurations, you should always use fully qualified domain names over short names."_

content/faq/validations/host-format.adoc
```diff
@@ -0,0 +1,20 @@
...
+
+If you want to have full support, change your host names to be FQDN.
```
JStickler: Full support for what?

I think what you want to say here is something more like, "For best results with Kiali, you should use fully qualified domain names when specifying hosts.

content/documentation/validations/index.adoc
xeviknal: My goal to say in that section is that: "Use FQDN or simple name, don't use service.ns format". So the implication you mention is correct. details preferred over details.bookinfo but details.bookinfo.svc.cluster.local is preferred over details.
Thanks for helping!

content/faq/validations/host-format.adoc
xeviknal: This is something copy-pasted from istio documentation. So I'd rather leave it as it is in the documentation, just in case.

content/faq/validations/host-format.adoc
xeviknal: Cool! :dancer:

PR #192, opened 2020-01-22
Blog posts not updated - fixing tests

xeviknal:
Latest blog posts sections wasn't updated automatically. It looks like the netifly engine was caching the html.

### DIFF
--- a/themes/kiali/layouts/partials/home/blog-latest-section.html
+++ b/themes/kiali/layouts/partials/home/blog-latest-section.html
@@ -1,8 +1,12 @@
   {{- $maxBlogPosts := .Site.Params.maxBlogPosts }}
-  {{- $json         := getJSON "https://api.rss2json.com/v1/api.json?rss_url=https://medium.com/feed/kialiproject" }}
+  {{/*expTimestamp: expire content every 1hour (6000s)*/}}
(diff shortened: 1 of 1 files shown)


israel-hdez:
I'll go ahead and merge to have the website updated.
But I don't understand. With this, will the website no longer need a manual rebuild when a new Medium post is published? or is this just preventing the Netlify engine to use outdated cached data when generating the website?

xeviknal:
Thanks @israel-hdez ! Because of the nature of hugo, a manual rebuild will still be necessary. The JSON fetch is done in building time.
Therefore, the change in the URL ensure that every time we re-build the netlify/rss2json.com engine will re-fetch the content.

jshaughn:
If I recall the API doc also requires a manual update.  I wonder if we could get that doc automatically updated as well.

israel-hdez:
@jshaughn Probably, an issue is necessary to have somebody assigned to address those two issues.

xeviknal:
After the latest blog post released by @israel-hdez the web page was not reloaded automatically. I can confirm the need of a deployment for that end. I think this is because of the nature of hugo, a static web page framework. The fetch of posts is done at build time.

PR #198, opened 2020-02-05
Sitemap.xml fix: need baseURL set to be compliant

xeviknal:
Google wants us to have a sitemap.xml with absolute urls.
Screenshot

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1,6 +1,7 @@
 languageCode = "en-us"
 title = "Kiali"
 theme = "kiali"
+baseURL = "https://kiali.io"
 
 [params]
     custom_css = [


abonas:
do we also have it submitted per the guidelines?
https://support.google.com/webmasters/answer/183668?hl=en#addsitemap

xeviknal:
I tried to submit it into the Google Search Console. The [current sitemap](https://kiali.io/sitemap.xml) is incomplete (only relative path) so the console didn't take it in. Once this PR is in, I will re-submit the [sitemap](https://deploy-preview-198--kiali.netlify.com/sitemap.xml) to the console.
We need the sitemap online deployed under kiali.io. It is not something we can upload as a file in the console. Moreover, the sitemap is dynamically updated in each deployment and google periodically comes to kiali.io/sitemap.xml to see updates.

israel-hdez:
The baseURL was removed in the past.

From history, I see it was removed because of the way the FAQ page is build. If you go to the preview of the FAQ (https://deploy-preview-198--kiali.netlify.com/faq/ ) all links on that page are built and redirect to the real website, rather than to the preview. When the FAQ page was introduced, this was making a little hard to develop the FAQ because you needed to adjust the URL to your local dev env.  

So, I'll merge because it is, probably, more important to have a correct sitemap file than the convenience of dev mode for the FAQ. It will be appreciated if you can do a follow-up PR to fix this little regression for a dev-friendly FAQ page.

PR #221, opened 2020-04-07
Versioned documentation

cfcosta:
This PR contains the script to generate and the generated archive for all Kiali versions since v1.0.0.

The matching of versions is done by #231.

## Running locally

You will need #231 to fill in the tags before doing anything else, then:

```
./scripts/build-archive.sh
```

The script will go through each tag, so it can not run if you have a dirty staged area on git.

## How to test

The test rendering of the site that netlify points all the links to the production instance, so you'll need to change the domain to the correct test one. This might be something fixable on another PR.

Just change `https://kiali.io` to `https://deploy-preview-221--kiali.netlify.app` after clicking a link :)

Alternatively, run it on your machine and the links should be correct.

To view the new pages, go to the "Documentation" page and select "Archive". All the versions should be there.

## Known Issues

* Versions are not sorted semantically. This is a weird consequence of our layout and happens on every index page.


cfcosta:
@jshaughn both are fixed.

PR #226, opened 2020-04-20
Fix broken links, add muffet target to help detect broken links

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/2676

There is now `make muffet` which runs a tool called "muffet" that checks for broken links.

The tool isn't 100% accurate (some links are really OK, but it reports them as bad) but it is better than nothing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ muffet:
 	@if ! curl -s --fail http://localhost:1313 > /dev/null ; then echo "You must start a Hugo local server!"; exit 1; fi
 	@mkdir -p $(CURDIR)/output/muffet
 	@ if [ ! -f $(CURDIR)/output/muffet/bin/muffet ] ; then GOPATH=$(CURDIR)/output/muffet go get -u github.com/raviqqe/muffet; fi
-	@$(CURDIR)/output/muffet/bin/muffet http://localhost:1313/
+	@$(CURDIR)/output/muffet/bin/muffet --exclude \#\!\(forum\|msg\)\/kiali-\(dev\|users\)\/? --exclude https?:\/\/localhost:20001\/? --exclude \/documentation\/developer-api http://localhost:1313/
 
 # Targets used to build and use the Hugo+Asciidoctor docker image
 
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -19,6 +19,13 @@ build:
 	sh scripts/build-swagger.sh
 	hugo
 
+# muffet - checks for broken links - you have to have hugo running for this to work
+muffet:
+	@if ! curl -s --fail http://localhost:1313 > /dev/null ; then echo "You must start a Hugo local server!"; exit 1; fi
+	@mkdir -p $(CURDIR)/output/muffet
+	@ if [ ! -f $(CURDIR)/output/muffet/bin/muffet ] ; then GOPATH=$(CURDIR)/output/muffet go get -u github.com/raviqqe/muffet; fi
+	@$(CURDIR)/output/muffet/bin/muffet http://localhost:1313/
```
brianredbeard, line 27: Still working through this a little more on my side, but it looks like adding these options take care of a number of the anchor problems:

`--exclude \#\!\(forum\|msg\)\/kiali-\(dev\|users\)\/? --exclude=https?:\/\/localhost:20001\/?`

jmazzitelli: I added those and one more additional one for that documentation/developer-api which appears to be some dynamically generated doc

jmazzitelli:
I made some final corrections. This should be ready for final review and approval/merge.

jmazzitelli:
Note that there are still some links that muffet for some reason doesn't like. Things like:

* https://github.com/kiali/kiali/blob/a660a80b2add1fd2fcfb5662c63824ca1dff95b9/operator/deploy/kiali/kiali_cr.yaml#L602-L608
* https://twitter.com/kialiproject
* https://github.com/kiali/kiali#grafana
* https://github.com/kiali/kiali#jaeger

and some others that are obvious examples in the docs and not meant to be true links like "http://grafana-server:3000/d/xyz/my-grafana-dashboard?var-namespace=some-namespace&var-app=some-app"

But again, its better than nothing.

xeviknal:
it makes sense to me. perhaps the next iteration for this is to exclude all these known URLs and then, prevent the deploy if there are URLs broken.

PR #233, opened 2020-05-13
Versioned documentation hotfixes

cfcosta:
Part of https://github.com/kiali/kiali/issues/2181.

This PR includes some commits that were not part of the merge, for some reason...

* Removes final .0 from versions, as we do not generate documentation for patch/snapshot versions.
* Use unordered lists instead of numbered ones for index pages.

End result with the changes:
image


cfcosta:
@jshaughn @israel-hdez hey, changed the urls, everything should be ok now.

PR #236, opened 2020-05-19
Change versioned documentation structure

cfcosta:
This PR adjusts the structure to be simpler to navigate and to make the distinction between current and older versions.

What I did:

* Removed the submenus from the Documentation tag, now that we need to choose the version.
* Pointed all the other links that are in the documentation to point to `latest`, which is a symlink to the latest released version (to be changed on the CI release process.
* Made all links point to either `latest` and or the latest released version.
* Added a new index page with some clarifications.
* Changed the swagger script to not fail under the new structure.

How that looks:
image

## FAQ

1. Where do I add new documentation?

`content/documentation/edge` is the documentation for master.


cfcosta:
When testing, please remember that because of some weird netlify thing, every link from an index page points to kiali.io instead of the test deployment. You just need to run locally or change the domain after clicking in a link.

cfcosta:
@jshaughn fixed the problem, can you take a look again?

cfcosta:
@mtho11 found the problem, am adding a fix.

@jshaughn that was on purpose, I think it was @jmazzitelli that pointed that when you are on the documentation for a version X, the masthead links should point to the version that you are currently on, not the latest one, and since that is not possible, I removed the links.

Another option would be to bring the menu items back, and point them to latest all the time, which is what we did before.

jshaughn:
> @jshaughn that was on purpose, I think it was @jmazzitelli that pointed that when you are on the documentation for a version X, the masthead links should point to the version that you are currently on, not the latest one, and since that is not possible, I removed the links.

OK, I need to pay better attention, this is the second time today I din't quite understand what was going on...  I can't say I like the fact that as designed, to even discover what documentation is available you need to first go to documentation, then pick a version.

Knowing this we may need to think a little more.  I would expect that until I explicitly select an archived version that I would see the latest (so right now, 1.18).  If I visit kiali.io and click Documentation I think I'd want to see, for the latest release:

image

And on that page also be given an option to get to an Archived version.

What do you think?
cc @mtho11 @jmazzitelli

jmazzitelli:
1. Go here: https://deploy-preview-236--kiali.netlify.app/documentation/v1.15/
2. See all the links on the bottom half (the ones under "Index") go to v1.15 - this is good.
3. Look at all the masthead links - all go to "master" - to me this is bad.

To me, as I click around, I want to stay on the version I am on (I picked v1.15 for a reason - I want to see the v1.15 documentation). If I want to see the latest docs, I'll explicitly do so by going to "Documentation" and selecting it. I don't like the big links at the top not taking me to the version of docs i'm already on.

cfcosta:
> 1. Go here: https://deploy-preview-236--kiali.netlify.app/documentation/v1.15/
> 2. See all the links on the bottom half (the ones under "Index") go to v1.15 - this is good.
> 3. Look at all the masthead links - all go to "master" - to me this is bad.
> 
> To me, as I click around, I want to stay on the version I am on (I picked v1.15 for a reason - I want to see the v1.15 documentation). If I want to see the latest docs, I'll explicitly do so by going to "Documentation" and selecting it. I don't like the big links at the top not taking me to the version of docs i'm already on.

That is what I don't get, to me it makes total sense that the links on the masthead always point to latest, because the masthead is a site-wide menu, not a contextual one, else it would be included in the same area as the content (like the breadcrumbs).

If you click on a link outside the content area then, imho, I'd expect it to be a global thing and not related to whatever old version I'm looking at.

jshaughn:
> That is what I don't get, to me it makes total sense that the links on the masthead always point to latest, because the masthead is a site-wide menu, not a contextual one

I agree with you @cfcosta , I think it's OK if the masthead always link to latest.  But I stand by my previous comment above, we can't hide the latest documentation under an archived version link.  The archive should be to get to older versions, not the current version.

cfcosta:
> > That is what I don't get, to me it makes total sense that the links on the masthead always point to latest, because the masthead is a site-wide menu, not a contextual one
> 
> I agree with you @cfcosta , I think it's OK if the masthead always link to latest. But I stand by my previous comment above, we can't hide the latest documentation under an archived version link. The archive should be to get to older versions, not the current version.

We could just revert the menu items change, and an "All versions" link there pointing to the documentation index, or just keeping the Documentation link poiting to the version selection.

jshaughn:
@cfcosta I don't know really.  Am I the only one that thinks it's bad to have a two step process to get to see the latest, or archived, documentation?   As a typical user I expect to just see the latest documentation when I visit a site.  To get to archived docs I want to be able to choose the version.  I don't expect to have to choose the latest version just to see what documentation exists.   Also, as presented it's not even clear what the latest release version is, I have to scan the long list and figure out that 1.18 is the highest version, and click that.  Maybe it's just me, would like to hear another opinion.

cfcosta:
@jshaughn in the page, the link to the latest version is on the first line, with relevant pages on the second. Would making it more noticeable better?

And I'd say that yeah, for the main documentation page I'd be ok with an intermediate page, but not for the other ones, like Features, Getting Started and so on on the masthead.

jshaughn:
> the link to the latest version is on the first line

I wasn't referring to "edge" but rather the latest release.  Actually, I think maybe "draft" may be better than "edge" and perhaps it should be separate, at the bottom.   Can decide later.

Anyway, Waiting for someone else to comment, anyone, because maybe I am off-base here :)

cfcosta:
> > the link to the latest version is on the first line
> 
> I wasn't referring to "edge" but rather the latest release. Actually, I think maybe "draft" may be better than "edge" and perhaps it should be separate, at the bottom. Can decide later.

Yeah, I meant latest, not edge:

image

And the "edge" name came from https://edgeguides.rubyonrails.org/ , but "draft" sounds good as well.

PR #251, opened 2020-06-23
Add FAQ entry for RBAC-enabled auth strategies

israel-hdez:
Fixes kiali/kiali#2595.


israel-hdez:
@jshaughn I'm not very, very happy with some wording. So I let you check it and see if something can improved :)

israel-hdez:
@jshaughn Thanks for the review. Merging.

PR #253, opened 2020-06-29
Add faq to versioned documentation

cfcosta:
Closes https://github.com/kiali/kiali/issues/2910


cfcosta:
So, at it turns out, https://github.com/kiali/kiali.io/pull/259 also fixes a part of this. Also added some the correct tags to get the layout (it was implied from the path, so it broke).

This is ready for review.

PR #255, opened 2020-07-04
Fix developer api sections

cfcosta:
Closes https://github.com/kiali/kiali/issues/2902.

The problem that is happening is because the generated developer API were being ignored and only generated on deploy, which now must be different because of the versioning. I'll be taking a look at retrofitting data to the older versions (so that we can properly point to `latest` on the menu) , but for now this should fix the problem.


cfcosta:
@jshaughn there is, I am on it.

PR #257, opened 2020-07-08
Quick install page

xeviknal:
Fixes https://github.com/kiali/kiali/issues/2251


content/documentation/edge/installation-guide/index.adoc
```diff
@@ -1,8 +1,9 @@
 ---
-title: "Getting started"
+title: "Installation guide"
 aliases:
+- /installation-guide/
```
cfcosta, line 4: Remember that this gets versioned, so after the next release there's a chance those are going to break. I'm not exactly sure how hugo handle more than one page with the same alias, this seems to be working now but it you plan to depend on this link, that's something you need to check.

content/documentation/edge/quick-install/index.adoc
```diff
@@ -0,0 +1,72 @@
...
+[source,bash]
+----
+istioctl dashboard kiali
+----
+
+icon:bullhorn[size=1x]{nbsp} The default credentials to log in are: `admin/admin`. It is highly recomendable to change them as soon as possible, see the link:{{< ref "/installation-guide#_login_options" >}}[Login Options]
+
+== Using Kiali Operator
+
+This option installs the Kiali Operator and the Kiali CR with default options. If you are running with an https://github.com/istio/istio/releases[upstream Istio release], you will probably want to pass in _--accessible_names="**"_ since that is the typical way Kiali is expected to run within an upstream Istio environment.
```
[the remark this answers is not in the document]
xeviknal: It needs to be changed in the getting-started page too then. It was copied from there. Good catch.

config.toml
```diff
@@ -36,8 +36,8 @@ relativeURLs = true
           parent = "main_news"
   [[menu.main]]
       identifier = "main_gettingstarted"
-      name = "Getting Started"
-      url = "/documentation/latest/getting-started"
+      name = "Quick install"
```
israel-hdez, line 39: May "Quick start" sound better? :)

config.toml
```diff
@@ -62,9 +62,9 @@ relativeURLs = true
              weight = 1
              parent="doc_features"
       [[menu.main]]
-          identifier = "doc_gettingstarted"
+          identifier = "doc_installation_guide"
           name = "Getting Started"
-          url = "/documentation/latest/getting-started"
+          url = "/documentation/latest/installation-guide"
```
israel-hdez, line 67: You probably should put a redirection on the old "getting-started" link.

I'm checking my old medium posts and in several posts I used the old (and now non-existent) https://kiali.io/documentation/getting-started/#_quick_install link. This is not your fault, but removing the `/documentation/latest/getting-started` page will *again* break external links.

In general, we should do a better job at not breaking external links. In particular, getting started or install pages are pretty important and we should try to never break potential external refferals.

xeviknal: Good point. I'll see what I can do. I see I can add alias however they are directed to the edge version. Have to get more familiar with the whole versioning system. :+1:

content/documentation/edge/installation-guide/index.adoc
xeviknal: TBH, I am no that familiar with the versioning system. I just replicated what it was before in the getting-started.
Is it worth it to add this new file under both `latest` and `edge`?

config.toml
cfcosta: You can direct the link to latest, probably on config.toml.

content/documentation/edge/installation-guide/index.adoc
cfcosta: latest is just a symbolic link to the latest released version, so the links could point there. "edge" means "latest master".

jshaughn: an aside, but I still don't care for 'edge'.  I would have preferred 'staging' or even 'master'.
.

cfcosta: @jshaughn I'm not married to the name at all, feel free to change :)

config.toml
xeviknal: yep, way better!

config.toml
xeviknal: I've put some alias to restore previous broken links and the one that will be broken bc of this PR.

content/documentation/edge/installation-guide/index.adoc
xeviknal: @cfcosta this aliases were already there before this PR. It looks like when there are duplicates, hugo takes the latest release. I am not 100% why takes this release and not another one. I've tried to add a v1.20, v1.21, v1.30 and in all the cases it has always taken the highest release, so it seams quite safe.

See the example: kiali.io/gettingstarted (been there since first release)

cfcosta: I'm pleasantly surprised :P Was not expecting it to work, but if it does, who am I to complain? :)

xeviknal:
@israel-hdez @jshaughn let me know if you have any other concern I should approach before we merge this. thx

config.toml
```diff
@@ -62,9 +62,9 @@ relativeURLs = true
              weight = 1
              parent="doc_features"
       [[menu.main]]
-          identifier = "doc_gettingstarted"
+          identifier = "doc_installation_guide"
           name = "Getting Started"
```
israel-hdez, line 66: Perhaps rename to "Installation guide" for consistency on referrals?

PR #258, opened 2020-07-10
Add kiali events on our home page

xeviknal:
I've added a new section in our home page where events related to kiali are going to be listed.

The source of this is a Google SpreadSheets which only the kiali team can edit. This spreadsheet is similar to the resulting one from the public form available to the community: bit.ly/3egKs58

If this PR gets merged, I'll invite to all of the kiali team to that spreadsheet.

wdyt?


themes/kiali/layouts/partials/home/events-latest-section.html
```diff
@@ -0,0 +1,43 @@
...
+          {{ $title   := index $r 1 }}
+          {{ $author  := index $r 4 }}
+          {{ $dateRaw := index $r 2 }}
+          {{ $date    := $dateRaw }}
+          {{ $summary := index $r 3 | plainify | truncate 150 }}
+
+          <a href="{{ $link }}" title="{{ $title }}" target="_blank">
+            <div>
+              <header>
+                <div>{{ $date }}</div>
```
israel-hdez, line 32: Can you use the same date format as in the blog posts cards?
This format can lead to confusion because depending on the country of the reader, entries change positions.

xeviknal: Good point! I gave up after some time playing w the limited pool of tools in hugo... however, I did manage to achieve the very same format. Thanks!

PR #259, opened 2020-07-14
Fix filtering on listing page

cfcosta:
This fixes an error on the layout that was filtering the index page when accessing it through the `latest` endpoint (accessing the version directly was ok, though).

This happens because Hugo does not know about symlinks and can not expand the url, unfortunately.

Before:
image

After:
image


cfcosta:
> @cfcosta I'll conditionally approve but I want to note that latest is showing 1.19 but the latest release is 1.20.  If that is expectedin the PR then OK, otherwise not.

@jshaughn this is expected, since the script has not had a chance to run automatically. Will create a PR later today.

jshaughn:
@cfcosta OK, yes, please gen the 1.20 version (only 4 days to 1.21).

PR #263, opened 2020-07-15
Fix FAQ Links

cfcosta:
Closes https://github.com/kiali/kiali/issues/2994

So, at it turns out, our faq being (barely) functional before the versioning is just an unlikely alignment between hugo and the content  we had at the time. As soon as things get more complicated it breaks down horribly.

This should fix some of the problems, at least the ones we've encountered until now.

* Fix header links.
* Add the content directly to the page.
* Fix some of the links on old versions of the "getting-started" page.

Proof of life:
image


jmazzitelli:
All these changes look valid.

PR #269, opened 2020-07-28
update sprint vids 42/41/40

jshaughn:


mtho11:
@jshaughn I was getting ready to do this as well ;)

mtho11:
@jshaughn I'm glad you found those videos because I was on PTO for both of those sprint demos. I know what I'll be watching...

PR #270, opened 2020-07-28
remove login and ldap auth strategies

jmazzitelli:
https://github.com/kiali/kiali/issues/3017


jmazzitelli:
> I don't think this is sufficient. I also see hits in quick-start.

where?

jmazzitelli:
I see nothing in quick-start: https://kiali.io/documentation/staging/quick-start/

jmazzitelli:
@jshaughn I think the problem was this wasn't rebased - you were looking at the old text. Check again - I just rebased on master.

PR #280, opened 2020-08-06
updates due to helm chart changes

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/3094

The Quick Start and Installation Guide pages are the two things that were changed.

### DIFF
--- a/content/documentation/staging/installation-guide/index.adoc
+++ b/content/documentation/staging/installation-guide/index.adoc
@@ -31,7 +31,7 @@ Kiali requires a supported version of Istio. link:https://istio.io/news/[The Ist
 |Notes
 
 | 1.7.0+
-| 1.22.0+
+| 1.22.1+
 | Istio 1.7 istioctl will no longer install Kiali. Use the Istio samples/addons all-in-one yaml or the Kiali Helm Chart for quick demo installs.
 
 | 1.6.0+
@@ -75,11 +75,11 @@ If you are going to install the Helm Chart, *you must use Helm 3*. You can insta
 
 == Installation
 
-Kiali can be installed in an OpenShift or Kubernetes cluster environment. For OpenShift only, a minimal preparation is necessary.
-
-icon:bullhorn[size=1x]{nbsp} If you previously installed Kiali through some other mechanism, *you must uninstall it first* using that original mechanism's uninstall procedures. There is no migration path between older installation mechanisms and the install mechanisms explaned below.
+icon:bullhorn[size=1x]{nbsp} If you previously installed Kiali through some other mechanism, *you must uninstall it first* using that original mechanism's uninstall procedures. There is no migration path between older installation mechanisms and the install mechanisms explained below.
 
 icon:lightbulb[size=1x]{nbsp}For several commands listed on this page, the OpenShift CLI command `oc` is used to interact with the cluster environment. If you are on Kubernetes, simply replace `oc` with `kubectl` unless otherwise noted.
+
+If installing Kiali into an OpenShift Kubernetes distribution then follow the minimal preparation below. For other Kubernetes distributions you can skip to the next section.
 
 
 === OpenShift Preparation
--- a/content/documentation/staging/installation-guide/index.adoc
+++ b/content/documentation/staging/installation-guide/index.adoc
@@ -73,16 +73,7 @@ For more, see the link:https://docs.okd.io/latest/welcome/index.html[OKD install
 If you are going to install the Helm Chart, *you must use Helm 3*. You can install Helm by following the link:https://helm.sh/docs/intro/install/[Helm installation docs].
 
 
-== Installation
-
-icon:bullhorn[size=1x]{nbsp} If you previously installed Kiali through some other mechanism, *you must uninstall it first* using that original mechanism's uninstall procedures. There is no migration path between older installation mechanisms and the install mechanisms explained below.
-
-icon:lightbulb[size=1x]{nbsp}For several commands listed on this page, the OpenShift CLI command `oc` is used to interact with the cluster environment. If you are on Kubernetes, simply replace `oc` with `kubectl` unless otherwise noted.
-
-If installing Kiali into an OpenShift Kubernetes distribution then follow the minimal preparation below. For other Kubernetes distributions you can skip to the next section.
-
-
-=== OpenShift Preparation
+=== OpenShift Requirements
 
 If you are installing on OpenShift, you must grant the `cluster-admin` role to the user that is installing Istio and Kiali. If OpenShift is installed locally on the machine you are using, the following command should log you in as user `system:admin` which has this `cluster-admin` role:
 
@@ -90,6 +81,13 @@ If you are installing on OpenShift, you must grant the `cluster-admin` role to t
 ----
   oc login -u system:admin
 ----
+
+icon:lightbulb[size=1x]{nbsp}For most commands listed on this page the Kubernetes CLI command `kubectl` is used to interact with the cluster environment. On OpenShift you can simply replace `kubectl` with `oc`, unless otherwise noted.
+
+
+== Installation
+
+icon:bullhorn[size=1x]{nbsp} If you previously installed Kiali through some other mechanism, *you must uninstall it first* using that original mechanism's uninstall procedures. There is no migration path between older installation mechanisms and the install mechanisms explained below.
 
 
 === Quick Install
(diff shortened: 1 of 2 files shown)


content/documentation/staging/installation-guide/index.adoc
```diff
@@ -32,7 +32,7 @@ Kiali requires a supported version of Istio. link:https://istio.io/news/[The Ist
 
 | 1.7.0+
 | 1.22.0+
```
jshaughn, line 34: May as well make this 1.22.1+

content/documentation/staging/installation-guide/index.adoc
```diff
@@ -68,10 +68,17 @@ For development the requirements are lower, depending on how demanding your appl
...
 
 
+=== Helm Requirements
+
+If you are going to install the Helm Chart, *you must use Helm 3*. You can install Helm by following the link:https://helm.sh/docs/intro/install/[Helm installation docs].
+
+
 == Installation
 
 Kiali can be installed in an OpenShift or Kubernetes cluster environment. For OpenShift only, a minimal preparation is necessary.
```
jshaughn, line 78: OpenShift is Kubernetes, no?  Maybe just say,
"If installing Kiali into the OpenShift Kubernetes distribution then follow the minimal preparation below.  For other Kubernetes distributions you can skip to the next section."

Also, move this below the bullhorn and lightbulb.

Also, if this is a valid requirement it would also be needed by the quick start, which users can navigate to directly.

content/documentation/staging/installation-guide/index.adoc
```diff
@@ -68,10 +68,17 @@ For development the requirements are lower, depending on how demanding your appl
...
 
+icon:bullhorn[size=1x]{nbsp} If you previously installed Kiali through some other mechanism, *you must uninstall it first* using that original mechanism's uninstall procedures. There is no migration path between older installation mechanisms and the install mechanisms explaned below.
```
jshaughn, line 80: explaned --> explained

content/documentation/staging/quick-start/index.adoc
```diff
@@ -16,108 +16,52 @@ weight: 2
 :imagesdir: /images/quickinstall/
 :sectlinks:
 
-icon:lightbulb[size=1x]{nbsp} This page is intended for users installing Kiali for the first time, not recommended for custom production environments. In case of custom installations, visit the link:{{< ref "/documentation/staging/installation-guide" >}}[installation guide].
+icon:lightbulb[size=1x]{nbsp} This page is intended for users installing Kiali for the first time. These instructions are not recommended to install Kiali in production environments. In the case of custom installations, visit the link:{{< ref "/documentation/staging/installation-guide" >}}[installation guide].
 
-== Installing Kiali
+== Install via Kiali Server Helm Chart
```
jshaughn, line 21: I think there needed to be a reference to the prerequistes section and also the OpenShift prep note before proceeding.  User should go there, make sure everything is kosher, and then return.

content/documentation/staging/installation-guide/index.adoc
jmazzitelli: done

content/documentation/staging/installation-guide/index.adoc
jmazzitelli: done.

content/documentation/staging/quick-start/index.adoc
jmazzitelli: done

content/documentation/staging/quick-start/index.adoc
```diff
@@ -16,108 +16,59 @@ weight: 2
 :imagesdir: /images/quickinstall/
 :sectlinks:
 
-icon:lightbulb[size=1x]{nbsp} This page is intended for users installing Kiali for the first time, not recommended for custom production environments. In case of custom installations, visit the link:{{< ref "/documentation/staging/installation-guide" >}}[installation guide].
+icon:lightbulb[size=1x]{nbsp} This page is intended for users installing Kiali for the first time. These instructions are not recommended to install Kiali in production environments. In the case of custom installations, visit the link:{{< ref "/documentation/staging/installation-guide" >}}[installation guide].
 
-== Installing Kiali
+icon:lightbulb[size=1x]{nbsp} If you are installing on OpenShift, you must grant the `cluster-admin` role to the user that is installing Istio and Kiali. If OpenShift is installed locally on the machine you are using, the following command should log you in as user `system:admin` which has this `cluster-admin` role: `oc login -u system:admin`
```
lucasponce, line 21: Please, OpenShift shoudn't be in the quickstart, it's a platform specific requeriment.

content/documentation/staging/quick-start/index.adoc
```diff
@@ -16,108 +16,59 @@ weight: 2
...
 
-=== Install via OperatorHub
+icon:lightbulb[size=1x]{nbsp} Before attempting to install Kiali, please review the link:../installation-guide/#_prerequisites[prerequisites].
```
lucasponce, line 23: For example, this link is +1, platform specific can go to prerequisites.

content/documentation/staging/quick-start/index.adoc
```diff
@@ -16,108 +16,59 @@ weight: 2
...
 
-This is the simplest install mechanism. If you are running OpenShift (or you have a Kubernetes environment with OperatorHub installed), simply go to the OperatorHub console and choose to install the Kiali Operator.
+== Install via Kiali Server Helm Chart
```
lucasponce, line 25: I suggest to use a previous point like 

```
== Install via Istio addons

From Istio 1.7 Kiali addon can be installed from:
$ kubectl apply -f https://raw.githubusercontent.com/istio/istio/release-1.7/samples/addons/kiali.yaml
```

Then, if I want to install via Helm is for two potential reasons:

a) I didn't find how to do that from the Istio page
b) I want to uninstall and install a new Kiali version different from the default one (and then I want to continue reading the helm install point).

content/documentation/staging/quick-start/index.adoc
jmazzitelli: well you and @jshaughn need to come to an agreement, because I added it as part of his review.

content/documentation/staging/quick-start/index.adoc
jmazzitelli: I do not like putting in version-specific instructions here. Once 1.8+ is released, these docs immediately become outdated and we'll have to remember to change these. 

Regarding this point:

> a) I didn't find how to do that from the Istio page

Instead of putting those instructions here, we should just have a link to that Istio page to help the person find it. That way we address two things: (1) we can help solve the "I didn't find it from the Istio page" problem and (2) we therefore do not have a copy or a repeat of the Istio docs.

content/documentation/staging/quick-start/index.adoc
lucasponce: QuickStart is not intended to OpenShift users.
Please, don't put it there.

OpenShift users have several ways to install Istio, to me install istio upstream in OpenShift is a corner case that a) it's described as a platform specific in Istio page b) should be also follow that format in the quick start.

QuickStart is for community users that want to focus on features.

content/documentation/staging/quick-start/index.adoc
lucasponce: To me, Istio.io and Kiali.io share this and it should be there.
Otherwise we force the new comers to understand the integration relation from the product and I don't think we want that.

To update this for 1.7 and 1.8 seems a very low effort but to have a one line to install things in the quickest way.

content/documentation/staging/quick-start/index.adoc
jshaughn: > QuickStart is not intended to OpenShift users.

I see no reason to ruin the quick start if the user is using an OS kubernetes env.  It's just one line.  Having said that, perhaps a decent compromise is to treat that OpenShift requirement as a re-requesite and move the OS note to that section, which is already mentioned in the quick start.

lucasponce: It's a prerequesite and as commented, the better place is the installation guide.
Istio has more than a dozen of platform specifics:
image

I'm okey to not specify them if we don't need to, but I don't think is good that it's marked in the Quick Start, it may create a biased perception that OpenShift is necessary (it had happen in the past).

content/documentation/staging/quick-start/index.adoc
jmazzitelli: we can add that - but realize there is no migration between that and our helm. So they will have to know to uninstall first before upgrading kiali via helm. I do have that warning, but of course i'm sure some people will not do it :)

jmazzitelli: OK. I added something that I think addresses all of our concerns.

PR #285, opened 2020-08-07
fix helm command and add a bullhorn for openshift 4.5 users

jmazzitelli:


jmazzitelli:
i forgot this is going in staging. we can merge it. it will go in the next rlease.

PR #289, opened 2020-08-11
Update validation with kiali-1471 issue

xeviknal:
Updating validation to https://github.com/kiali/kiali/issues/1471

### DIFF
--- a/content/documentation/staging/validations/index.adoc
+++ b/content/documentation/staging/validations/index.adoc
@@ -958,7 +958,7 @@ A VirtualService defines a set of traffic routing rules to apply when a host is 
 
 ==== Resolution
 
-This is valid configuration only if VirtualService is bound to a gateway, sidecars do not accept this behavior. There are several caveats when using this method and defining the same parts in multiple Virtual Service definitions is not recommended. While Istio will merge the configuration, it does not guarantee any ordering for cross-resource merging and only the first seen configuration is applied (rest ignored). As recommended, each VS definition should have a 'catch-all' situation, but this can only be defined in a definition affecting the same host.
+This is a valid configuration only if two VirtualServices share the same host but are bound to a different gateways, sidecars do not accept this behavior. There are several caveats when using this method and defining the same parts in multiple Virtual Service definitions is not recommended. While Istio will merge the configuration, it does not guarantee any ordering for cross-resource merging and only the first seen configuration is applied (rest ignored). As recommended, each VS definition should have a 'catch-all' situation, but this can only be defined in a definition affecting the same host.
 
 ==== Severity
 
@@ -1107,5 +1107,3 @@ include::/data/files/validation_examples/304.yaml[]
 ==== See Also
 
 https://github.com/kiali/kiali/tree/master/business/checkers/common/workload_selector_checker.go[Validator source code, window="_blank"] +
-
-'''


content/documentation/v1.22/validations/index.adoc
```diff
@@ -958,7 +958,7 @@ A VirtualService defines a set of traffic routing rules to apply when a host is
 
 ==== Resolution
 
-This is valid configuration only if VirtualService is bound to a gateway, sidecars do not accept this behavior. There are several caveats when using this method and defining the same parts in multiple Virtual Service definitions is not recommended. While Istio will merge the configuration, it does not guarantee any ordering for cross-resource merging and only the first seen configuration is applied (rest ignored). As recommended, each VS definition should have a 'catch-all' situation, but this can only be defined in a definition affecting the same host.
+This is valid configuration only if two VirtualServices share the same host but are bound to a different gateways, sidecars do not accept this behavior. There are several caveats when using this method and defining the same parts in multiple Virtual Service definitions is not recommended. While Istio will merge the configuration, it does not guarantee any ordering for cross-resource merging and only the first seen configuration is applied (rest ignored). As recommended, each VS definition should have a 'catch-all' situation, but this can only be defined in a definition affecting the same host.
```
lucasponce, line 961: "This is valid configuration" -> "This is a valid configuration"

xeviknal: Approached! Also moved this to staging, not to latest.

israel-hdez: I was reviewing this as part of the medium sprint post and the first sentence is somewhat confusing or it's contradicting itself.

It reads `This is valid configuration only if two VirtualServices share the same host but are bound to a different gateways` but in the [Istio best-practices docs](https://istio.io/latest/docs/ops/best-practices/traffic-management/#split-virtual-services), the provided example is binding more than one VS to the same GW. So, unless I'm misunderstanding, this sentence is probably not right at all.

Also, when looking at the whole sentence, it reads `This is valid configuration only if [...], sidecars do not accept this behavior`. I find it somewhat contradictory.

Probably, it's a wording issue? I'm guessing it should be "This is valid configuration ~~only if~~ **unless** two VirtualServices share the same host but are bound to a different gateways, sidecars do not accept this behavior."

xeviknal: Thanks for pointing at this one.

Regarding to the behavior, I guess that we wanted to show a warning when different VS share the same host. Not to point that the configuration is wrong, but to point out to users that this might be a source of potential problems. This is because of merging mechanism: if there are duplicated routes, regexps that might compete one to the other, then the behavior might be unexpected and no-deterministic.

On the other hand, we could put in place a validation with error severity to specifically point this statement `A VirtualService can only be fragmented this way if it is bound to a gateway.` [1]

Regarding to the wording, I guess that the previous description was more accurate that the one in place for this PR. The meaning is: The Virtual Service you have in place is invalid if there isn't any gateway attached. The merger doesn't apply to this kind of virtual services.
Plus, I would add a clarification saying that this warning is just to point a potential source of problems.

wdyt? 

[1] https://istio.io/latest/docs/ops/best-practices/traffic-management/#split-virtual-services

israel-hdez: Well, my point is that the sentence is confusing as it is right now (or at least I find it confusing). I just wanted to note that this probably needs a small change to remove confusion :)

About adding an extra validation, I think it will appreciate it.

PR #290, opened 2020-08-11
Add a 'Getting Started Video' section 

mtho11:
This adds a 'Getting Started Videos' so that the 'Latest videos' section can be just that the latest cool videos; but that we have a highly visible place for beginners to see how to use Kiali.

fixes: https://github.com/kiali/kiali/issues/3117


mtho11:
CC: @lucasponce @xeviknal

PR #296, opened 2020-08-27
Add new video content to getting started and latest videos sections.

mtho11:
Add new video content to getting started and latest videos sections.

- Helm chart install into Istio 1.7.x
- Enable Istio Auto injection from Kiali

### DIFF
--- a/themes/kiali/layouts/partials/home/video-getting-started-section.html
+++ b/themes/kiali/layouts/partials/home/video-getting-started-section.html
@@ -10,6 +10,7 @@
     </h1>
 
     <div class="embed-video-links">
+        <iframe class="video-iframe-links" src="https://www.youtube.com/embed/ht9M_GW7k6c"></iframe>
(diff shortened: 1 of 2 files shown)


themes/kiali/layouts/partials/home/video-getting-started-section.html
```diff
@@ -10,6 +10,7 @@ <h1>
     </h1>
 
     <div class="embed-video-links">
+        <iframe class="video-iframe-links" src="https://www.youtube.com/embed/ht9M_GW7k6c"></iframe>
```
jshaughn, line 13: Not a blocker, but not sure if we want to limit these two three videos to avoid wrap on the website.

mtho11: Ok good idea. I'm doing one more video. I will incorporate that for that video.

PR #298, opened 2020-08-29
Add new logging enhancements video.

mtho11:
Add new logging enhancements video to kiali.io

### DIFF
--- a/themes/kiali/layouts/partials/home/video-latest-section.html
+++ b/themes/kiali/layouts/partials/home/video-latest-section.html
@@ -10,6 +10,7 @@
     </h1>
 
     <div class="embed-video-links">
+        <iframe class="video-iframe-links" src="https://www.youtube.com/embed/Elj1r5qzre8"></iframe>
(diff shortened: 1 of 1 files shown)


lucasponce:
Videos are starting to be packed in the page:

image

There 13 videos on the main page, which looks to much.

I think it's better to put the latest (2,3 or 4) and also perhaps organize the youtube channel.

Or to link the videos with the features, but let's not populate more these sections.

But if we want to add more than 2,3 or 4, we can use a scroll section similar as the blog, otherwise user won't reach that after the videos there are more sections.

mtho11:
Yes, @lucasponce I had the same thought. The next PR will be to add horizontal scrolling to the video sections. That sound like a good solution to you?

lucasponce:
> Yes, @lucasponce I had the same thought. The next PR will be to add horizontal scrolling to the video sections. That sound like a good solution to you?

+1 as it's already introduced in the blog post section, it sounds good.

jshaughn:
If possible, add titles to the videos as well, it is not always possible to know what the video is from the youtube title.

mtho11:
I created https://github.com/kiali/kiali/issues/3164 for the above comments.

PR #300, opened 2020-08-31
Enhance video sections to be more like blog section.

mtho11:


mtho11:
I just did it for the Latest Videos section to get approval 

video-scroll

mtho11:
@jshaughn @lucasponce  WDYT?

lucasponce:
The idea of the scroll is good.

But the horizontal space between videos is not same as the blogs/events cards.

Also the fonts used and icon looks different than the cards, perhaps if those can be align as well it would look closer.

From the Blog list:
image

From the Event list:
image

mtho11:
Ok, I think I have managed to make all the style changes consistent:

latest-videos2

lucasponce:
> LGTM. Not a blocker for me but if you can left-align align the Blogs and the Events sections with the videos that would be even nicer.

+1

lucasponce:
I've merged to progress but I couldn't test locally.

image

I've realized that at the moment we use a wide resolution, the kiali.io starts to look bad.

@mtho11 would you mind to send a followup fix for this ?

mtho11:
@lucasponce following up on fix

mtho11:
Looks like we also need a fix for the features section as it doesn't align as well...

Kiali__Service_mesh_observability_and_configuration

Someone want to tackle that one?

israel-hdez:
The features card are supposed to be centered on the page.

Indeed, everything is supposed to be centered on the page. It's just that widths vary and some blocks have same with.

mtho11:
Ah ok, thanks @israel-hdez

PR #302, opened 2020-09-03
Health configuration documentation

aljesusg:


lucasponce:
@aljesusg Please also include this feature in the new Features page:

https://kiali.io/documentation/latest/features/

Take a look how we are doing with similar sections (i.e. Validations), so, this feature should be listed there and the specific page can be linked from there to expand/detail the config details, but we want to keep consistency in the doc in the Features.

There are other pages that are little bit outdated and needs to be reviewed in the same style as well.

lucasponce (changes requested):
This should be consolidated in the Features section, too.

lucasponce:
Adding @jshaughn, he makes good revision on the doc as well, both English language and topic.

content/documentation/staging/features/index.adoc
```diff
@@ -529,6 +529,28 @@ Kiali uses Istio Wizards to generate Istio Traffic config for a specific Service
...
+<div style="display: flex;">
+ <span style="margin: 0 auto">
+  <a class="image-popup-fit-height" href="/images/documentation/health-configuration/beta.png" title="beta Namespace">
+   <img src="/images/documentation/health-configuration/beta.png" style="width: 300px;height:400px;display:inline;margin: 0 auto;" />
+  </a>
+  <a class="image-popup-fit-height" href="/images/documentation/health-configuration/details_health.png" title="Health Deatails">
+   <img src="/images/documentation/health-configuration/details_health.png" style="width: 300px;height:400px; display:inline;margin: 0 auto;" />
+  </a>
+ </span>
+</div>
```
jshaughn, line 550: This screenshot should be updated given the late changes to the popup content.

jshaughn (changes requested):
I am going to prepare a sub-PR rather than edit via comments.  But see the note about the screenshot.

update: https://github.com/aljesusg/kiali.io/pull/4

content/documentation/staging/features/index.adoc
aljesusg: totally I did the screenshot before the last discussion :) Thanks @jshaughn

jmazzitelli:
I don't know if this is technically related, but it sort of is - what about documenting the settings for "system" health configuration - i.e. all the component status checks that we recently changed? We need that documented somewhere, perhaps it can go in this PR somewhere?

I bring this up because someone yesterday didn't know about the new stuff and was also confused - even I am not 100% sure how all of this is configured now. See: https://github.com/kiali/kiali/issues/3201

lucasponce:
@jmazzitelli I think your comment is valid, but perhaps it's off topic of this work.

But I agree that configuration relies on the Kiali CR only and probably the "Config" section:

image

should be updated / improved for easy refer.

Probably not all topics, it may have a generic one and specifics for important features that deserves more detail.

content/documentation/staging/features/index.adoc
```diff
@@ -529,6 +529,28 @@ Kiali uses Istio Wizards to generate Istio Traffic config for a specific Service
 </div>
 ++++
 
+== Health Configuration
+
+Kiali allows users to customize the configuration used to determine health.
```
lucasponce, line 534: This deserves more wording.
Like the health is a composition of several inputs, pods, traffic, etc.
Global health is calculated like the max of all inputs and Health Configuration allows to determine the health of a particular signal.

On the status of the feature, traffic health is supported.

Re-arrange my wording, but I guess it can give you an idea about what I mean.

content/documentation/staging/configuration/health/index.adoc
```diff
@@ -0,0 +1,384 @@
...
+toc::[]
+:toc-title: Custom Health
+:keywords: Kiali Documentation Health
+:icons: font
+:imagesdir: /images/documentation/health-configuration/
+
+
+== Health
+
+Customize resource health by overriding the default health configuration.
```
lucasponce, line 24: Again, please don't assume the user knows how health works.
It requires an intro, that you will see pods / traffic and probably more signals in the future.

A global health is calculated as the max severity of all inputs/signals.

This feature allows to determine the severity of those signals.

content/documentation/staging/configuration/health/index.adoc
```diff
@@ -0,0 +1,384 @@
...
+  health_config:
+    rate:
+      - namespace: ".*"
+        kind: ".*"
+        name: ".*"
+        tolerance:
+          - code: "^5\\d\\d$"
+            direction: ".*"
+            protocol: "http"
+            degraded: 0.1
```
lucasponce, line 72: Why don't leave this to 0 ?
To have a 0.1 is really odd to explain.
Perhaps I missed this in the code, but I suggest to adjust this.

aljesusg: It was in 0.1 % for me is the same to set it to 0 or let it....In the code I just checking if error_rate is > 0 and > degraded value so it's going to work with 0. It backend we  have this values so we should change them if we want to keep this aligned

jshaughn (commented):
@aljesusg The additional description and table is good.  I suggest just we modify the terminology a little bit.  See https://github.com/aljesusg/kiali.io/pull/5.

jshaughn (approved):
LG

lucasponce (approved):

PR #303, opened 2020-09-10
[Fix] video alignment issues with other sections at certain resolutions.

mtho11:
A fix to my last PR that fixes the misalignment at some resolutions. I didn't realize that some resolutions are fixed pixels and some (>1500) are 15% padding.


jshaughn (approved):
Looks great, let's merge if you don't have anything else pending.

PR #310, opened 2020-09-23
- Add relnotes v1.24

jshaughn:
- Add sprint 45 sprint demo
  - update sprint demo titles with associated version

### DIFF
--- a/content/news/release-notes.adoc
+++ b/content/news/release-notes.adoc
@@ -52,7 +52,6 @@ Fixes:
 * https://github.com/kiali/kiali/issues/3199[Broken navigation within details page tabs]
 * https://github.com/kiali/kiali/issues/3202[Configurable Tresholds - 0% is shown in Failure objects]
 * https://github.com/kiali/kiali/issues/3198[Links (sort of) broken when navigating within details page tabs]
-* https://github.com/kiali/kiali/issues/3187[Add new config `rev` into kiali settings]
 * https://github.com/kiali/kiali/issues/3172[latest kiali.io docs are still 1.22 when they should be 1.23]
 * https://github.com/kiali/kiali/issues/3110[travis produced a test-race failure]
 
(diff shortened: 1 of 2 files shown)


content/news/release-notes.adoc
```diff
@@ -15,6 +15,47 @@ For additional information check our https://www.youtube.com/channel/UCcm2NzDN_U
...
+* https://github.com/kiali/kiali/issues/3015[Better link Istio configuration under Workload and Service details.]
+
+Fixes:
+
+* https://github.com/kiali/kiali/issues/3181[istio_status_enabled not working in 1.23.0]
+* https://github.com/kiali/kiali/issues/3171[Traces view: sometimes not easy to click on traces]
+* https://github.com/kiali/kiali/issues/3199[Broken navigation within details page tabs]
+* https://github.com/kiali/kiali/issues/3202[Configurable Tresholds - 0% is shown in Failure objects]
+* https://github.com/kiali/kiali/issues/3198[Links (sort of) broken when navigating within details page tabs]
+* https://github.com/kiali/kiali/issues/3187[Add new config `rev` into kiali settings]
```
jmazzitelli, line 55: This should not be here. We changed the way this is implemented - there is no longer a "rev" or "revision" config setting. It's [the "config_map_name" now.](https://github.com/kiali/kiali-operator/blob/v1.24.0/roles/default/kiali-deploy/defaults/main.yml#L137) - see https://github.com/kiali/kiali/pull/3206 and https://github.com/kiali/kiali-operator/pull/127

jmazzitelli (changes requested):

content/news/release-notes.adoc
jshaughn: the way to fix this in general is that if an issue gets closed that should not be mentioned in the relnotes, it should not be in the sprint "Done" column (i.e. it needs to be removed from the sprint project)

jmazzitelli (approved):

PR #318, opened 2020-10-15
Documentation for OpenID with RBAC disabled

israel-hdez:
Adding documentation to configure Kiali with OpenID authentication and
RBAC disabled.

Related kiali/kiali#3084


israel-hdez:
@sandhose You may want to review this, because it is adding the missing JWT checks you mentioned in https://github.com/kiali/kiali/pull/2798#issuecomment-636168758.

In this, users can opt to not configuring OIDC integration in Kubernetes and still be able to use OpenId in Kiali, but this requires a full validation of the JWT token.

I hope this is doing it correctly. Your review will be appreciated.

cfcosta (approved):
Looks good to me.

PR #319, opened 2020-10-22
add new FAQ section named "Installation" with first entry

jmazzitelli:
Add FAQ to educate people about the error that will occur if you use clusterRoleCreator=false and accessible_namespaces of **.

We will need this since we needed to rip out the human-readable error due to https://github.com/kiali/kiali-operator/pull/167

Note this introduces a new section of the FAQ named "Installation"


jshaughn (approved):
LGTM.  If the contect is immediately usable you could also make the same changes to /latest but it's fine as is if we can wait until next Friday.

PR #323, opened 2020-10-29
Remove version 1.0 as we no longer support it.

mtho11:
This PR just removes the 1.0 version of the documentation as we no longer support it. 

Version 1.12 is now our oldest supported documentation.


jshaughn (approved):
LGTM

PR #336, opened 2020-12-18
Add documentation for usage of OpenId with Keycloak

cfcosta:


content/documentation/staging/configuration/authentication/openid.adoc
```diff
@@ -274,3 +274,11 @@ spec:
       http_proxy: http://USERNAME:PASSWORD@10.0.1.1:8080/
       https_proxy: https://USERNAME:PASSWORD@10.0.0.1:8080/
 ----
+
+=== Usage with Keycloak
```
jshaughn, line 278: Maybe change to "Using Keycloak" for consistency with the section headers above.

jshaughn (changes requested):
@cfcosta looks fine to me, although I'll add @israel-hdez to further validate the content.  if this is applicable to the 1.28 release please also duplicate the addition so that it shows up in 'latest'.

cfcosta:
@jshaughn thanks for the input!

content/documentation/staging/configuration/authentication/openid.adoc
```diff
@@ -274,3 +274,11 @@ spec:
...
       https_proxy: https://USERNAME:PASSWORD@10.0.0.1:8080/
 ----
+
+=== Using with Keycloak
+
+When using OpenId with Keycloak, you will need to enable the `Standard Flow Enabled` and `Implicit Flow Enabled` options on the Client (in the Administration Console):
+
+image::/images/documentation/authentication/keycloak-implicit-client.png[Client configuration screen on Keycloak]
+
+Without those, you will get an error like `Client is not allowed to initiate browser login with given response_type. Implicit flow is disabled for the client.`.
```
israel-hdez, line 284: Yes, but no.

In the `Requirements` section, the first paragraph says:

```
If you want to enable usage of the OpenId’s authorization code flow, make sure that the Kiali’s signing key is 16, 24 or 32 byte long. If you setup a signing key of a different size, Kiali will only be capable of using the implicit flow.
```

So, it depends on user config. If the user configures a signing key of any of the mentioned lengths, the `Implicit Flow Enabled` option is not really required. Moreover, its safer to leave it disabled, but it depends on user needs.

I think this paragraph will be confusing to some people. Because in the _Introduction_ section it's said that `Kiali supports the authorization code flow (preferred) and the implicit flow`, and IMO this paragraph is implicitly telling users that Kiali needs "Implicit flow" for auth to succeed :/

israel-hdez (commented):
The content itself is fine, but I'm a little reluctant to merge, because correctness is not good enough.

You need either the `Standard` or the `Implicit`. It depends on the length of the Kiali signing key.

jshaughn (commented):
@cfcosta +1 on adding 1.28, make sure you update both with @israel-hdez suggested changes.

cfcosta:
@jshaughn @israel-hdez thanks for the feedback! I'm reading a little bit more, see how to make this less confusing.

jshaughn:
@cfcosta Let's get this updated and in ASAP, also now we need to duplicate for 1.29 as well.  If there is an issue please work with @israel-hdez  to get proper content.

cfcosta:
@israel-hdez can you take a look and see if the new wording makes more sense?

israel-hdez (approved):
Quite better.

PR #338, opened 2020-12-21
docs: Tidy up some of the English in Badly Laid Out Graph FAQ

asummers:
Happy to rework any other verbiage while I'm here, but this seemed to be the least invasive way of editing this particular paragraph.


cfcosta:
Looks great, thanks for the pull request. Will merge.

More invasive changes are also welcome, as long as the end result is better. Might need some feedback from the author in some cases, but always welcome.

PR #341, opened 2021-01-18
Faster site generation + selinux compat

jotak:
- Add a new make target (serve-latest) to speed up site generation by
  skipping all archived versions
- Add ":z" flag for selinux compatibility on volume mounts

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,5 +26,16 @@ build-hugo:
 .PHONY: serve
 serve: build-hugo
 	@mkdir -p ./resources/_gen/images && mkdir -p ./resources/_gen/assets
-	@${CONTAINER_RUNTIME} run -t -i --sig-proxy=true --rm --mount type=bind,src=$(shell pwd),dst=/site -w /site -p 1313:1313 kiali/hugo:${HUGO_VERSION} /hugo serve --baseURL "http://localhost:1313/" --bind 0.0.0.0 --disableFastRender
+	@${CONTAINER_RUNTIME} run -t -i --sig-proxy=true --rm -v "$(shell pwd)":/site:z -w /site -p 1313:1313 kiali/hugo:${HUGO_VERSION} /hugo serve --baseURL "http://localhost:1313/" --bind 0.0.0.0 --disableFastRender
+	@rm -rf ./resources
+
+# This will ignore old versioned documentation for serving, to speed up the generation process
+serve-latest: build-hugo
+	@mkdir -p ./resources/_gen/images && mkdir -p ./resources/_gen/assets && mkdir -p ./.ignore
+	# move away all "v*" directories, that is, all versions except "staging".
+	@mv  ./content/documentation/v* ./.ignore
+	# move back just the version that stands for latest
+	@mv  "./.ignore/$(shell basename $(shell readlink -f content/documentation/latest))" ./content/documentation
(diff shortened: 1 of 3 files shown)


jotak:
Hi,
I've noticed the build time is very long due to all the archived versions (like 1 minute on my machine), so here's a new make target for a faster generation.

Also the serve target wasn't working on my machine and it took me a while to understand that, once again, that was due to selinux (wink @jmazzitelli ).

It would be nice to check by people who use docker and/or without a system with selinux

lucasponce:
The idea of this PR is good as the standard "serve" is slow.

But also I'd say that the "normal" process to write new doc is to write on the "staging" area, latest if the the published one.

Perhaps this new make can focus on both latest and staging.

I agree that modifying doc for old versions is rare and can get triggered on purpose only

jotak:
> The idea of this PR is good as the standard "serve" is slow.
> 
> But also I'd say that the "normal" process to write new doc is to write on the "staging" area, latest if the the published one.
> 
> Perhaps this new make can focus on both latest and staging.

Yes that's actually what it does, it keeps both latest and staging. This is what I've written in the readme: `[...] Instead of generating everything, you can speed up the process if you're only interested in latest or staging version.`

> 
> I agree that modifying doc for old versions is rare and can get triggered on purpose only

Makefile
```diff
@@ -26,5 +26,15 @@ build-hugo:
...
-	@${CONTAINER_RUNTIME} run -t -i --sig-proxy=true --rm --mount type=bind,src=$(shell pwd),dst=/site -w /site -p 1313:1313 kiali/hugo:${HUGO_VERSION} /hugo serve --baseURL "http://localhost:1313/" --bind 0.0.0.0 --disableFastRender
+	@${CONTAINER_RUNTIME} run -t -i --sig-proxy=true --rm -v "$(shell pwd)":/site:z -w /site -p 1313:1313 kiali/hugo:${HUGO_VERSION} /hugo serve --baseURL "http://localhost:1313/" --bind 0.0.0.0 --disableFastRender
+	@rm -rf ./resources
+
+# This will ignore old versioned documentation for serving, to speed up the generation process
+serve-latest: build-hugo
+	@mkdir -p ./resources/_gen/images && mkdir -p ./resources/_gen/assets && mkdir -p ./.ignore
+	@mv  ./content/documentation/v* ./.ignore
+	# move back just latest
+	@mv  "./.ignore/$(shell basename $(shell readlink -f content/documentation/latest))" ./content/documentation
```
lucasponce, line 37: Perhaps this line confused me, I haven't yet tested locally, but "latest" link doesn't point to "staging" folder, but last released version.

jotak: what it does is copying all "v*" folders to ".ignore/", except the one that corresponds to "latest" link (which is put back to its place as a second step). The "staging" folder doesn't match "v*" so it is not moved.

jotak: let me add a comment

cfcosta (approved):
I don't use SELinux, but can not comment on that. However, I did run serve-latest and it is working, can verify that only staging and latest are there :)

lucasponce:
I was testing locally, I found a minor thing but perhaps not to worry:

```
Press Ctrl+C to stop
^Cmake: [Makefile:39: serve-latest] Error 255 (no tiene efecto)
```

When stopping the local server with Ctrl+C I had that message but I didn't see any major issue in my local branch.

jotak:
@lucasponce yes, killing the server with ctrl+c apparently triggers an error return code of the process, which would make the Make target fail and stop processing. I think it happens also on the existing target `make serve`, the difference is that I'm telling make here to ignore the error so that it continues processing the target, to run the cleaning steps (this is done by prefixing a command with `-`)

PR #342, opened 2021-01-18
Fix minimum privileges docs

israel-hdez:
Doc says that minimum privileges required to login will turn kiali
unusable. This is not totally true, as pages that work purely from
metrics will work OK. So, changing that paragraph to be more accurate.

Fixes kiali/kiali#3575


jshaughn (approved):
LOK

PR #351, opened 2021-03-03
Add tool to check built site

israel-hdez:
Added tool is htmlproofer: https://github.com/gjtorikian/html-proofer

The main goal is to find any broken links in the resulting website.
However, the tool is capable of doing other checks.

Adding the verification as a CI check in GitHub Actions.

Fixes kiali/kiali#2688


israel-hdez:
I think this is worth to have. I see lots of errors pointing broken links and missing media.

israel-hdez:
@lucasponce I'm wondering if you could check what's the right link on this line: https://github.com/kiali/kiali.io/blame/master/content/documentation/staging/tutorial/04-observe/03-details.adoc#L75

It's marked as broken (and it is broken).

israel-hdez:
All bad URLs are, apparently, fixed.
See the "green run" here: https://github.com/israel-hdez/kiali.io/pull/2/checks?check_run_id=2080332559

xeviknal:
@israel-hdez I like the idea of having a proofer in each PR. It will ensure that everyone opening a PR will also make sure to check if there are broken URLs. 

Couple of ideas to make this a bit less over-killing:
1) Only analyze the N last releases instead of all of them (like the supported ones 1.18+).
2) Does it work to just check for 4xx responses (I see there is the `--only-4xx` flag or `--http-status-ignore`). Because there are some site which change the folder architecture but keeping the link working. e.g. [google forums urls](https://github.com/kiali/kiali.io/pull/351/files#diff-3f14b833f11127fb2ebad51b0650252d4e4cb050d83789f25e7736647658df9fL39-L40) (although they are redirected the original URLs work).

What do you think?

israel-hdez:
> Couple of ideas to make this a bit less over-killing:
> 
> 1. Only analyze the N last releases instead of all of them (like the supported ones 1.18+).

I'm both :+1: and :-1: ... 

So, I'm :+1: because I think it's too much effort to maintain older versions of docs. And it aligns with our policy about only supporting the latest Kiali version.

But then, I'm :-1: because, if we don't want to maintain older docs, what's the point on keeping those older versions?

Honestly, the only reason I chose to fix older docs, is because I hate broken links in websites ;-)... But then, most of the time the broken link was also present in newer versions of docs, it was a matter of replicating the fixes with `patch` commands.

I'd seek for advice from others about this. If we choose to not worry about older docs, I think we can use the `file_ignore` flag.
Anyway, I'd like this to move forward while we discuss with the rest of the team. I can do a follow-up PR if consensus is to not validate the older docs.

> 2. Does it work to just check for 4xx responses (I see there is the `--only-4xx` flag or `--http-status-ignore`). Because there are some site which change the folder architecture but keeping the link working. e.g. [google forums urls](https://github.com/kiali/kiali.io/pull/351/files#diff-3f14b833f11127fb2ebad51b0650252d4e4cb050d83789f25e7736647658df9fL39-L40) (although they are redirected the original URLs work).
> 

I think it depends a lot on how the redirection is done. In the GForums case, apparently, the redirection is done with Javascript which gives us a false positive of a `hash does not exist` error (see https://github.com/israel-hdez/kiali.io/runs/2071590408#step:7:188).

There were lots of `hash does not exist` errors. This one is from the tutorial which is quite recent: https://github.com/israel-hdez/kiali.io/runs/2061233448#step:7:338. Most of the time, these errors were benign: the page may still exist, but the link is broken because the section within the page does not exist anymore. So, I'm "-1" on silencing `hash does not exist` errors.

xeviknal:
> But then, I'm ￼ because, if we don't want to maintain older docs, what's the point on keeping those older versions?

@israel-hdez you are right. Completely

> Anyway, I'd like this to move forward while we discuss with the rest of the team.

I agree with that. Let's give this a run and see what are the reactions. I find this very useful overall. 
Plus I am eager to see what happen with redirections.

Good job!

xeviknal (approved):
Increasing standards in our website.
In favor of giving this a long run and tweak it over the time.
LGTM

jshaughn:
> I agree with that. Let's give this a run and see what are the reactions. I find this very useful overall.
Plus I am eager to see what happen with redirections.

+1, let's get this in.   After that we can see how it goes.  If we need to increase the perf and want to limit checks to newer versions, I am fine with that.  In fact, I'd be fine wil just checking staging and latest, as long as any issues found were appropriately patched in any of the older versions we sill support.   Also, limiting to 404 is fine.  But if it's fast enough then I don't see why we should put any more effort into this, and just scan all versions for all problems.

israel-hdez:
Thanks for the reviews!
I'm merging.

PR #365, opened 2021-04-21
OpenID docs for Azure

israel-hdez:
Fixes kiali/kiali#3294


israel-hdez:
Hi, @kulbhusanazad

This is adding a small section to document how to use AKS-managed AAD with Kiali. I hope you can validate this.

---
@xFayre, @tesharp, @thomasmRavn, this may also be of your interest.

kulbhusanazad:
Hi @israel-hdez 

below are my comments on this

**1. In the AD Client App** : issuance of the access token needs to be selected at least , sample screen shot below

image




**2. In the Kiali CR**  :

i)
issuer_uri: `https://sts.windows.net/<tenant-id>/`

ii) It would be good to include the secret needed for accessing the AD application since you mentioned about its creation. So if someone is following this literally then would know the secret is also needed in the kiali cr

tesharp:
Regarding issuer_uri it will be different if using a v2 token. Its https://sts.windows.net/<tenant-id>/ for v1 tokens, but v2 is https://login.microsoftonline.com/<tenant>/v2.0

xFayre:
This is awesome! thanks @israel-hdez

kulbhusanazad:
> Regarding issuer_uri it will be different if using a v2 token. Its https://sts.windows.net// for v1 tokens, but v2 is https://login.microsoftonline.com//v2.0

i think passing the resource parameter is only valid for v1 and not for v2.

israel-hdez:
> > Regarding issuer_uri it will be different if using a v2 token. Its https://sts.windows.net// for v1 tokens, but v2 is https://login.microsoftonline.com//v2.0
> 
> i think passing the resource parameter is only valid for v1 and not for v2.

I would like to use what it's known to work. So, I changed it to the "sts" url :-) 
@tesharp @kulbhusanazad if you can confirm that also the "login" url will work OK, we can put a note later that either issuer_uri could be used.

---

@kulbhusanazad I added your image and a bullet telling to make sure that "access tokens" is enabled.

About the secret, it shouldn't be needed to put it in the Kiali CR. It isn't read from there. It's even not mentioned in the set of available parameters for openid (see https://github.com/kiali/kiali-operator/blob/7920dddefb84e3b32231c89b0d826f1bbc0b0ddb/deploy/kiali/kiali_cr.yaml#L144-L158). I'm quite sure that the secret should be created only as a Kubernetes secret.


(same link to view the updated draft)

kulbhusanazad:
> > > Regarding issuer_uri it will be different if using a v2 token. Its https://sts.windows.net// for v1 tokens, but v2 is https://login.microsoftonline.com//v2.0
> > 
> > 
> > i think passing the resource parameter is only valid for v1 and not for v2.
> 
> I would like to use what it's known to work. So, I changed it to the "sts" url :-)
> @tesharp @kulbhusanazad if you can confirm that also the "login" url will work OK, we can put a note later that either issuer_uri could be used.
> 
> @kulbhusanazad I added your image and a bullet telling to make sure that "access tokens" is enabled.
> 
> About the secret, it shouldn't be needed to put it in the Kiali CR. It isn't read from there. It's even not mentioned in the set of available parameters for openid (see https://github.com/kiali/kiali-operator/blob/7920dddefb84e3b32231c89b0d826f1bbc0b0ddb/deploy/kiali/kiali_cr.yaml#L144-L158). I'm quite sure that the secret should be created only as a Kubernetes secret.
> 
> (same link to view the updated draft)

Yes , sts(v1) url works for me but the v2 url does not work for me.

right , the secret i was mentioning is not openid parameter , rather its auth parameter in the kiali cr. My point was to have a mention or note about it as it should be created as kubernetes secret and have it referenced in the kiali cr for the open id to work completely. i have seen this has been mentioned in someplace of kiali doc. if you think this would be redundant information for this particular openid doc then its fine.

israel-hdez:
@kulbhusanazad The part I don't understand is when you say that the secret should be referenced in the KialiCR.

How are you referencing it?? Because there is no setting for it in the CR... Unless you are talking about this config? https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml#L342-L347

kulbhusanazad:
> @kulbhusanazad The part I don't understand is when you say that the secret should be referenced in the KialiCR.
> 
> How are you referencing it?? Because there is no setting for it in the CR... Unless you are talking about this config? https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml#L342-L347

Yes right, this is what i meant

israel-hdez:
@kulbhusanazad I added the following text in the second bullet:

```
Please, note that the suggested name for the Kubernetes Secret is kiali. If you want to customize the secret name,
you will have to specify your custom name in the Kiali CR. See the comments for the secret_name configuration
in the sample Kiali CR.
```

It has links to the relevant lines in the sample yaml file. The preview page should be updated.

Does this work for you?

kulbhusanazad:
> @kulbhusanazad I added the following text in the second bullet:
> 
> ```
> Please, note that the suggested name for the Kubernetes Secret is kiali. If you want to customize the secret name,
> you will have to specify your custom name in the Kiali CR. See the comments for the secret_name configuration
> in the sample Kiali CR.
> ```
> 
> It has links to the relevant lines in the sample yaml file. The preview page should be updated.
> 
> Does this work for you?

Yes , Perfect !!

israel-hdez:
@nrfox adding you as a reviewer, because I know you have good grammar.

nrfox (commented):
Content looks great! Just a few very minor suggestions.

israel-hdez:
@nrfox I integrated your feedback :)
I think you need to approve so that I can merge.

nrfox (approved):

israel-hdez:
@nrfox Thanks for your review.

@kulbhusanazad Thanks for all your help in making this possible.

PR #370, opened 2021-05-05
Add upgrade section and install docs on kiali-server chart

nrfox:
Adds an `Upgrade` section to the install guide that gives instructions for updating Kiali during an Istio canary upgrade and for upgrading Kiali itself. Also adds information in the install section for using the `kiali-server` helm chart to install Kiali without the operator.

Relates to https://github.com/kiali/kiali/issues/2884

### DIFF
--- a/content/documentation/staging/installation-guide/index.adoc
+++ b/content/documentation/staging/installation-guide/index.adoc
@@ -275,13 +275,17 @@ In general, Kiali tries to maintain compatability between at least the two most 
 
 For https://istio.io/latest/docs/setup/upgrade/canary/[Canary upgrades]:
 
-. Deploy the upgraded https://istio.io/latest/docs/setup/upgrade/canary/#control-plane[istio control-plane].
+. Deploy the upgraded https://istio.io/latest/docs/setup/upgrade/canary/#control-plane[Istio control-plane].
 . Update the Kiali Server configuration to point to the new Istio deployment. These fields need to be updated:
 - `spec.external_services.istio.config_map_name` to the new Istio configmap revision.
 - `spec.external_services.istio.istiod_deployment_name` to the new istio deployment revision.
 - `spec.external_services.istio.istio_sidecar_injector_config_map_name` to the new istio sidecar injector configmap revision.
 
-If you are using the *kiali-operator*, update the Kiali CR configuration:
+How you update these fields depends on how you have deployed Kiali.
+
+===== Operator
+
+If you are using the kiali-operator, update the Kiali CR configuration:
 
 [source,bash]
 ----
@@ -292,7 +296,9 @@ If you are using the *kiali-operator*, update the Kiali CR configuration:
 
 Wait until the operator restarts the Kiali Server and then verify everything is working correctly.
 
-If you are using the `kiali-server` *Helm Chart*, set the Helm Chart values: 
+===== Helm Chart - Kiali Server
+
+If you are using the `kiali-server` Helm Chart, set the Helm Chart values: 
 
 [source,bash]
 ----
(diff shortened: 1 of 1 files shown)


jmazzitelli:
Here's the netlify link to the page that was changed: https://deploy-preview-370--kiali.netlify.app/documentation/staging/installation-guide/

content/documentation/staging/installation-guide/index.adoc
```diff
@@ -253,6 +265,63 @@ icon:lightbulb[size=1x]{nbsp} If you installed the Kiali CR in a different names
...
+[source,bash]
+----
+  kubectl rollout restart deployments kiali -n istio-system
+----
+
+=== Upgrading Kiali
+
+==== Operator
+
+When upgrading Kiali using the operator, it's recommended to keep the operator version in sync with the Kiali version. The best way to do this is to ensure that your Kiali CR's `spec.deployment.image_version` field is set to `operator_version` (the default if it is unset) then upgrade the operator version. When the operator upgrade completes, the operator will upgrade the Kiali Server image version to match its own image version.
```
jmazzitelli, line 319:
```suggestion
When upgrading Kiali using the operator, it's recommended to keep the operator version in sync with the Kiali version. The best way to do this is to ensure that your Kiali CR's `spec.deployment.image_version` field is set to `operator_version` (this is the default if this value is left unset) then upgrade the operator version. When the operator upgrade completes, the operator will upgrade the Kiali Server image version to match its own image version.
```

content/documentation/staging/installation-guide/index.adoc
```diff
@@ -253,6 +265,63 @@ icon:lightbulb[size=1x]{nbsp} If you installed the Kiali CR in a different names
...
+  helm upgrade \
+    --set external_services.istio.config_map_name=istio-canary \
+    --set external_services.istio.istio_sidecar_injector_config_map_name=istio-sidecar-injector-canary \
+    --set external_services.istio.istiod_deployment_name=istiod-canary \
+    --repo https://kiali.org/helm-charts \
+    --namespace istio-system \
+    kiali-server kiali-server
+----
+
+Restart the Kiali pod to pickup the new configmap changes.
```
jmazzitelli, line 308: I don't know how to make this flow better, but I can only tell you what I first thought when I read this. I thought this was saying you should restart the pod regardless of whether you were using the server helm chart or operator.  Now, I know the answer :) but the flow of this made it ambiguous. Do I restart the pod even if I'm using the operator? It isn't really clear (though, yes, this appears after the sentence "If you are using the kiali-server".... but because there are no clear delineated sections here, it seems ambiguous.)

Perhaps all we need to do is change this sentence to be:
```suggestion
If you are using the `kiali-server` helm chart, you must restart the Kiali pod in order to pickup the new ConfigMap changes. This is not required if you are using Kiali Operator.
```

jmazzitelli (changes requested):

content/documentation/staging/installation-guide/index.adoc
nrfox: That's good feedback. I originally had *kiali-operator* and *Helm Chart* as subsection headers underneath the Canary section. Do you think that'd make it clearer? Something like

### Canary
For Canary Upgrades ...

#### Kiali Operator
If you are using the *kiali-operator*, update the Kiali CR configuration:
...

#### Helm Chart - Kiali Server
If you are using the `kiali-server` *Helm Chart*, set the Helm Chart values: 
...

I also wonder if it'd be helpful to call the kiali server helm chart deployment something like the "standalone" deployment or if that's more confusing?

jmazzitelli: I was thinking having separate sections like that. That's fine.

"standalone" deployment would be more confusing. I would keep using "Kiali Server Helm Chart" to refer to this.

jmazzitelli (approved):
this looks good to me

PR #372, opened 2021-05-11
Fixo typo in upgrade Kiali section

abonas:


aljesusg (approved):
LGFM

aljesusg:
> @aljesusg for some reason in netlify preview I see the old text and not the one that I fixed. I'm probably missing something

Mmm. I did a checkout and works. Maybe is cached in netlify. Can we force the build again?

nrfox (approved):
:+1:

nrfox:
> @aljesusg for some reason in netlify preview I see the old text and not the one that I fixed. I'm probably missing something

I don't see it under `/latest` but I do see it under `/staging`: https://deploy-preview-372--kiali.netlify.app/documentation/staging/installation-guide/#_upgrade which I believe is fine

abonas:
@nrfox what's the difference between staging/latest? both should be fixed?

jshaughn:
@abonas In netlify you sometimes have to manually update the link to see the correct version of the file.   /latest is just a synonym for our most recent release.  /staging is basically "Head" and is where you typically want to add new content as it will be copied to the release version during the the next release CI (which will then also become the new /latest).   If you don't add content to staging then it will only ever be present in the specific version for which it was committed.

aljesusg:
> > @aljesusg for some reason in netlify preview I see the old text and not the one that I fixed. I'm probably missing something
> 
> I don't see it under `/latest` but I do see it under `/staging`: https://deploy-preview-372--kiali.netlify.app/documentation/staging/installation-guide/#_upgrade which I believe is fine

Yes I saw it too. I thought that maybe Alissa had something cached in her browser

PR #390, opened 2021-07-06
Update compatibility chart.

jshaughn:
Kiali 1.35 adds support for WorkloadGroup, which requires istio 1.8.

Fixes https://github.com/kiali/kiali/issues/4154


jmazzitelli (approved):

abonas:
perhaps it makes sense to tweet about compatibility updates (or mention it with release)? @xeviknal @jshaughn wdyt?

jshaughn:
> perhaps it makes sense to tweet about compatibility updates (or mention it with release)? @xeviknal @jshaughn wdyt?

I defer to @xeviknal but given that this is an update old, out-of-support Istio versions, I'm not sure it's very exciting.

PR #395, opened 2021-07-26
Add Virtual Machine config info

nrfox:
Adds a page to the configuration section for setting up Kiali to work with VM based workloads.


content/documentation/v1.37/configuration/vm.adoc
```diff
@@ -0,0 +1,43 @@
+---
```
nrfox, line 1: Wasn't exactly sure where to put this info. Also considered putting it under FAQ since it's less about configuring Kiali and more about configuring Prometheus in order to get the telemetry from the VMs for Kiali to display.

nrfox:
I'm not sure why the link is 404ing in CI and causing a build failure. It seems like the external link should be either localhost or netlify and not `kiali.io`:

```
Checking 780 external links...
- ./public/documentation/staging/configuration/index.html
Ran on 759 files!
  *  External link https://kiali.io/documentation/staging/configuration/vm/ failed: 404 No error

HTML-Proofer found 1 failure!
```

content/documentation/staging/configuration/vm.adoc
```diff
@@ -0,0 +1,43 @@
...
+:toc-title: In this section:
+:keywords: vm virtual machine
+:icons: font
+:sectlinks:
+
+toc::[]
+
+== Introduction
+
+Kiali supports visualizing Virtual Machine workloads (WorkloadEntry) on the graph the same as pod-based workloads running inside a Kubernetes cluster. You must ensure that Istio Proxy is running and correctly configured on the Virtual Machine and Prometheus must be able to scrape the metrics endpoint of the Istio Proxy running on the VM. From there, Kiali will be able to read the traffic telemetry for the Virtual Machine workloads and incorporate the workloads into the graph.
```
jshaughn, line 18: suggested rewording:

Kiali graph visualizes both Virtual Machine workloads (WorkloadEntry) and pod-based workloads, running inside a Kubernetes cluster.  You must ensure that the Istio Proxy is running, and correctly configured, on the Virtual Machine.  Also, Prometheus must be able to scrape the metrics endpoint of the Istio Proxy running on the VM.   Kiali will then be able to read the traffic telemetry for the Virtual Machine workloads, and incorporate the VM workloads into the graph.

content/documentation/staging/configuration/vm.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+icon:bullhorn[size=1x]{nbsp} Kiali does not currently distinguish between pod-based and VM-based workloads nor does Kiali support viewing additional details for the VM-based workloads beyond what is displayed on the graph. One way to distinguish between the two is to give the VM-based workloads a different version label than the pod-based workloads.
+
+== Configuring Prometheus to scrape VM-based Istio Proxy
+
+Once the Istio Proxy is running on a Virtual Machine, configuring Prometheus to scrape the VM's Istio Proxy metrics endpoint is the only configuration Kiali needs to display traffic for the VM-based workload. link:https://prometheus.io/docs/prometheus/latest/configuration/configuration/[Configuring Prometheus] will vary between environments. Here's a very simple example of a Prometheus configuration that includes a job to scrape VM based workloads:
```
jshaughn, line 24: Here's  -->  Here is

jshaughn (changes requested):
I think putting this in Configuration is OK.   We are going to look soon at a re-org of content and maybe then we'll move things around.  A couple of edit suggestions.

Also, the bad link is because, I think, you need to make a change to config.toml to add the new entry into the overall structure.

nrfox:
> Also, the bad link is because, I think, you need to make a change to config.toml to add the new entry into the overall structure.

Updated `config.toml` but that made the link checker very angry. It looks like it's expecting the link at `/latest`. Do I need to add the doc to `/latest` in addition to `/staging?`

jshaughn (changes requested):
@nrfox Changes look good.   Right, so I guess changing config.toml was not the answer to the previous link issue :-)  Still, I think structurally, we need that change.  Because config.toml is not versioned, but global, we need to have those references resolved to `latest`.  So, I guess we need to add a /latest version of vm.adoc.  If the same info applies to 1.37 just copy the staging file.  If it doesn't apply make whatever changes you see fit. Thanks.

jshaughn (approved):
LGTM, I think these link issues may just be a catch 22 when adding new section like this.  We'll see... I'm merging.

nrfox:
> LGTM, I think these link issues may just be a catch 22 when adding new section like this. We'll see... I'm merging.

Yeah git won't let me add `/latest` since it's linked to v1.37

PR #397, opened 2021-08-04
document how to add a sidecar to kiali

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4258

netlify link: https://deploy-preview-397--kiali.netlify.app/documentation/staging/faq/#sidecar-injection

see also: https://github.com/kiali/kiali-operator/pull/388

### DIFF
--- a/content/documentation/staging/faq/installation/sidecar-injection.adoc
+++ b/content/documentation/staging/faq/installation/sidecar-injection.adoc
@@ -1,0 +1,14 @@
+---
+title: How can I inject an Istio sidecar in the Kiali pod?
+weight: 4
+---
+:icons: font
+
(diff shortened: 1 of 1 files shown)


jmazzitelli:
I confirmed that setting in the Kiali Cr does inject the sidecar.

The verify error is due to some links in the v1.34 docs that are no longer valid. We can ignore these for the purposes of this PR

nrfox:
I see the label on my kiali pod after editing my Kiali CR but the sidecar didn't actually get injected. When I do the namespace label the sidecar injection works. Maybe something with my env?

jmazzitelli:
> When I add the injection label to my namespace the sidecar injection works. Maybe something with my env?

Did you have canary setup or the namespae label with it disabled? Those seem to override the pod label. See:
https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy

nrfox:
> Did you have canary setup or the namespae label with it disabled? Those seem to override the pod label. See:
> https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy

Doesn't look like it but something else must be up with my env. The label is correct and matches what's in the istio docs you linked.

nrfox (approved):
+1 this is good to have documented

PR #401, opened 2021-08-17
WIP Split Features in different pages

lucasponce:
Fixes https://github.com/kiali/kiali/issues/2494

@xeviknal I have prepared a draft PR for this split.

The content is not updated, just focused on the "split".
A main intro page:
image

That put every feature/section into a different page.

I like the split in a first step as it may help that every section grows in a better direction, also there are other outdated section (CustomDashboards, Tracing) that should be consolidated in both features and configuration.

Also I have an open question if we may want to add the sub-menu into the features (similar as "Config" and "Glossary" sections):


xeviknal:
I think I like the idea of having the subsections into each category. I feel that this is more self-explenatory and guides better the users to see what's in kiali:

image

However, I do like the features section in the home page as well:
Screenshot of Kiali_ Service mesh observability and configuration
Perhaps we could reuse that section and place it in this features page. This should be the director to the main areas. Instead we'd have 7 pages instead of 17. In each of the 7 pages we could have the index with sub-indexes (as of we have today in the features list) specific of the area.

Just thinking out loud.

content/documentation/staging/features/01-istio-service-mesh-overview.adoc
```diff
@@ -0,0 +1,14 @@
+---
+title: "Service Mesh Overview"
+date: 2021-07-26T00:00:00+02:00
+draft: false
+weight: 1
```
xeviknal, line 5: I think that as long as you have the weight, you don't need to put the numerics index as a prefix int the file name.
The file name in hugo is used to shape the URL of the page. IMO, for better search engine practices, we could do without the prefix.

lucasponce: ah, right, I wonder if I can modify the url but not the file, I put the numeric index in the file to keep them ordered in the folder, but yes, the url is not needed, it's more to not get lost when working with long files.
I'll search if both aspects can be met.

lucasponce:
> Perhaps we could reuse that section and place it in this features page. This should be the director to the main areas. Instead we'd have 7 pages instead of 17. In each of the 7 pages we could have the index with sub-indexes (as of we have today in the features list) specific of the area.

I see what you mean. I also like the icons with the high level features. But I think there is no a clean 1:1 mapping between features icons and the content. Also this PR is not trying to re-visit the content itself but the structure.

There is a more generic issue about docs to revisit if the current content is valid.

Also, if we are not 100% sure, we can put this one on-hold until the main revisit issue is covered, but I think that split the features in pages would help us to maintain them better, and we can join some pages that today are added in the "documentation" menu but I think they don't follow the logic of the structure.

xeviknal:
Perhaps the illustrations aren't the best fit for this scenario. However, as you have it now, without sub-sections, sounds a bit empty.
Perhaps we can maintain the same index table as we have today but link it to the new files. It might need to work a bit with Hugo to understand how to process the sections in different files, but I think it is feasible.

jshaughn:
I do think we need to split up the features into more discrete pages.  I do wonder whether we actually need a "Features" section and maybe instead all of the content should just be in the "Documentation".  I think, like istio.io, maybe our Documentation menu item should not be a dropdown and just go to a single page that leads you to all the docs.

lucasponce:
> I do think we need to split up the features into more discrete pages. I do wonder whether we actually need a "Features" section and maybe instead all of the content should just be in the "Documentation".

+1 on that, I had also that impression that "Features" and "Documentation" were some sort of redundant and could be unified under a single "Documentation" layer.

Perhaps then this PR should be closed and re-arranged once this is discussed, it was more experimental.

jshaughn:
> Perhaps then this PR should be closed and re-arranged once this is discussed, it was more experimental.
> 
> But in general, I'm not happy with the "template", these kind of changes are hard and time consuming, that was reason I opened [kiali/kiali#4293](https://github.com/kiali/kiali/discussions/4293) as well.

I missed #4293 but I'm +1 to that sort of restructure, if it's doable without massive effort.  Then Documentation can just all be there.  My fear is just that we don't have the expertise to do it.  As for this PR, I'm not sure to close it or to use it to start in a consolidated Documentation structure.

lucasponce:
> As for this PR, I'm not sure to close it or to use it to start in a consolidated Documentation structure.

Yes, this or a new one works for me, then to summarize if we understand the same actions to perform:
- Move "Features" entry in the menu under a "Features" entry just in the "Documentation" menu.
- "Features" would be a landing page to split individually each feature into its own page.
- Under the "Documentation" menu, there are some entries that belongs to specific feature page, that could be re-arrange them as well.

@jshaughn does that sound ok ?

jshaughn:
> Move "Features" entry in the menu under a "Features" entry just in the "Documentation" menu.

I am +1 to removing the `Features` entry in the top menu, but actually, I was also suggesting we remove the change the `Documentation` menu as well, replacing the dropdown with an initial doc page (more like istio.io).

lucasponce:
I think we could unify the template/layout to something like this:
image

In the left, the user would always see the list of the pages, in the right, there would be a list of items for the current page.
The center would be used for the content.

In theory, this layout should be automatically generated on new pages and new content, or that should be the goal (not sure how to do it, but "it should be doable" ).

Then, only the main page should be different, and rest of the pages should try to follow this pattern to simplify the maintenance of the site.

If it sounds good, I can try to investigate if it's something that can be achieved, in theory, content should be reused.

lucasponce:
Oh, GH closed this PR when I reset it locally.
Anyway, it was a draft, I keep the discussion and I investigate about the structure and split of pages.

PR #409, opened 2021-08-31
[kiali.io] Improve social icons on footer and add some to header

jshaughn:
Fixes https://github.com/kiali/kiali/issues/3835

Footer:
- re-order icons by importance
- add spacing between and below
- add tooltips
- 
Header
- a couple of tweaks for better rendering of the search form.

### DIFF
--- a/themes/kiali/layouts/partials/mastfoot.html
+++ b/themes/kiali/layouts/partials/mastfoot.html
@@ -9,27 +9,38 @@
 
   <!-- Social icons -->
   <section id="footer-social-icons">
-    <a href="https://medium.com/kialiproject" target="_blank"
-      class="fab fa-medium-m">
-    </a><a href="https://github.com/kiali" target="_blank"
+    <a href="https://github.com/kiali" title="Kiali project on Github" target="_blank"
       class="fab fa-github">
-    </a><a href="https://twitter.com/kialiproject" target="_blank"
-      class="fab fa-twitter">
-    </a><a href="https://www.youtube.com/channel/UCcm2NzDN_UCZKk2yYmOpc5w" target="_blank"
-      class="fab fa-youtube">
-    </a><a href="https://istio.slack.com/archives/CJVRXB09F" target="_blank"
+    </a>
+    &nbsp;
+    <a href="https://istio.slack.com/archives/CJVRXB09F"  title="Kiali channel on Istio Slack" target="_blank"
       class="fab fa-slack">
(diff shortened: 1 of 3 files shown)


lucasponce (changes requested):
I think the style is not correct, see the screenshot
image

Perhaps you can put it together as part of the "Search" element to keep them in the same layout position ?

jshaughn:
> Perhaps you can put it together as part of the "Search" element to keep them in the same layout position ?

Hmm, I thought I tested that narrow-width scenario and the icons were above the options.  I'll re-check.  To be honest, I spent many hours trying to get something to work, the css here is perhaps beyond my abilities and tying to "Search" was my first attempt, and although it seemed promising, it just would not play nicely.  Very frustrating.  I'll go back and try again when I have a clear mind for it.

jshaughn:
@lucasponce I tried again, please take a look and see if it's better.

abonas:
looks good. perhaps (and only if this is not causing another 2 hour attempt) increase the icons just a bit?
Now I wonder how many users will scroll down to see all the rest of the icons...

lucasponce:
There is a minor style issue:
image

That makes those icons out of the page in normal resolutions.

My suggestion: let's remove the search box, today is useless any search can be done by Google (that box is just a workaround to the google query, it's not performing any query to any internal database), and let's use the "Search" space to place the icons.

jshaughn:
@abonas @lucasponce 

After wasting another 1/2 day on adding social icons to the masthead, I've reverted it and am giving up.  It may be easy, but for me I think it would require tearing down the whole masthead and building it up again. I can't figure the way to get it inserted, the logic is too tricky as we have all sorts of things going on in there wrt screen sizes, the generated nav, etc.  To be honest, I didn't think we needed them anyway as they are already in the footer and not they look a bit better there.

As for the search, I'd elect to keep it (over adding social icons), I actually like the constrained search.  With or without it I still had issues with the icons.

Finally, for whatever reason I couldn't increase the size of the footer icons, using the fas-2x (for example) seemed to increase the background but not the icon.

I invite someone else to take this one further, or we can take it as-is for now, and kick the can down the road.

lucasponce (approved):

lucasponce:
Thanks @jshaughn.
I know the effort and the problems it takes to manipulate the "template".
I'll merge to consolidate, as we have several PRs open that would touch the structure and template probably we can iterate on them.

PR #410, opened 2021-09-01
Update Tutorial to 1.39

jshaughn:
Part of kiali#4118


jshaughn:
@lucasponce , Here is a pass at updating the tutorial to 1.39.  One thing I did not do is augment the new screen shots with "focus-ovals".  I found them a little distracting and in many cases unncessary.  But, if you feel like we need them we can go back and add circles where you think it's important.

In some places I made small additions, changes or corrections.  I also added some information for OpenShift users and tried to use the same style as is there already for Minikube.

I know we originally targeted 1.36 as the version, to be in line with SM 2.1, but the delta between 1.36 and 1.39 isn't too huge and I felt this better served the [larger] upstream community.

lucasponce (approved):
This updates are very good.
I had an ultra minor comment that perhaps in the "Connect > Request Routing" I'd choose a Workload Graph with no Services to show better the "round-robin" between "travel-portal" and "travel-agency".

But just to comment something that can be addressed later, first this is good to merge.

PR #411, opened 2021-09-07
fix docs

haoxins:


content/documentation/staging/faq/authentication/token-strategy.adoc
```diff
@@ -11,7 +11,7 @@ You can extract a service account's token from the secret that was created for y
 For example, if you want to log into Kiali using Kiali's own service account, you can get the token like this:
 
 ```
-kubectl get secret -n istio-system $(kubectl get sa kiali-service-account -n istio-system -o jsonpath={.secrets[0].name}) -o jsonpath={.data.token} | base64 -d
+kubectl get secret -n istio-system $(kubectl get sa kiali-service-account -n istio-system -o jsonpath="{.secrets[0].name}") -o jsonpath={.data.token} | base64 -d
```
jmazzitelli, line 14: this change is not needed. I see no reason why the double quotes are required.

jmazzitelli (changes requested):
I do not see why this change is needed or required. The docs are correct. If this command fails, please post the error you see when you run it without the quotes. Otherwise, the docs should remain as is.

content/documentation/staging/faq/authentication/token-strategy.adoc
haoxins: 企业微信截图_b40f3e8c-bb11-4d7a-80a5-4a210b3b983c

jmazzitelli: Then it appears the issue is the command does not work when you use the "zsh" shell. This was not clear because you made no mention of "zsh" in the title of the PR (all you said was "fix docs" in the title) and you have no description whatsoever and no error messages in this PR.

Please update the title of this PR and also add a description. When we approve PRs, the PR must indicate why the PR is needed (i.e. what is this fixing?). It is best to create a [kiali/kiali issue](https://github.com/kiali/kiali/issues/new/choose) but for small issues like this in the docs, having a PR is fine but you need to explain why you are making this change. Just saying "fix docs" with absolutely no description in the PR is not helpful. We have no idea what this is trying to address. A person using "bash" will not see this error. This is a "zsh" specific fix which is also compatible with "bash" and other shells.

jmazzitelli: So... you refused to simply change the title and add a description before we merge this? That makes no sense. But OK.

PR #413, opened 2021-09-09
Split features page

lucasponce:
Fixes https://github.com/kiali/kiali/issues/2494

WIP: Most of changes consolidated in "latest" folder, but "stage" needs to be updated.


jshaughn (commented):
First impressions:
- I like the cleaner masthead with less options and the clean left/right split.  I don't think I will miss the search but I have used it in the past.
- I like the shorter Documentation dropdown
  - I'd prefer no dropdown and just a Documentation landing page, but this is OK.
- I'm OK with the Documentation secondary dropdowns (Features/Configuration/Glossary)
  - I'd prefer no secondary dropdowns and just a landing page for each
- I don't like the third "Hero" button for "Learn more about Kiali" as it's unclear how it differs from the other choices and that it is a video.  I'd prefer it move back to the "About" blurb or just be removed completely, in favor of the bottom videos, the quick start, tutorial, the features, etc.
- I don't like the reduced space between "About" and "Features", it's too squished.
- I like the removal of "out-of-support" versions
  - If/when we have branches this will not be as much of an issue but right now the old versions slow things down considerably when building.
- I like the Feature breakup
  - I don't think we even need the boilerplate text at the top of the Features page
  - The straight feature list is OK but this could probably be nicer

That's all I have for a first impression

lucasponce:
Thanks for the review.

   
>   * I'd prefer no dropdown and just a Documentation landing page, but this is OK.
> * I'm OK with the Documentation secondary dropdowns (Features/Configuration/Glossary)
>   
>   * I'd prefer no secondary dropdowns and just a landing page for each

I see what you mean, perhaps we can move to this when we put in place a global index in the left, typically sites allow to see a global structure of the documentation, the current template relies on the dropdowns.

But I guess that would trigger the discussions about changes in the template, for this PR I'd suggest to continue with the existing approach and let's divide the work of manipulating the non-content aspects of the site.

> * I don't like the third "Hero" button for "Learn more about Kiali" as it's unclear how it differs from the other choices and that it is a video.  I'd prefer it move back to the "About" blurb or just be removed completely, in favor of the bottom videos, the quick start, tutorial, the features, etc.

Ok, I was not fully convinced as well, I'll remove it.

> * I don't like the reduced space between "About" and "Features", it's too squished.

Ok, I see a different marging depending of the width, I'd adjust it.

>   
>   * I don't think we even need the boilerplate text at the top of the Features page

Yes, it looks like a repetition of the "About" main intro. I'll remove it.

>   * The straight feature list is OK but this could probably be nicer
> 

I tried to bring all the discussion with @xeviknal, now it's linked with the structure of features of the "home page".
Note that I've modified the "Metrics" icon to support the "Details" section.

Content has not been modified, I also think that perhaps the structure can be improved and provide a different approach as this one is bringing the original structure.

But at least the aim of this PR would be to split the features in individual pages and allow that the structure grows in this way. Then every contributor can easily maintain a section.

I'll add a new commit with the suggestions.

lucasponce:
Note that HTML-Proofer errors are expected
image

As it's comparing existing links in kiali.io which changes on them.

So, I wonder if we want to keep it as part of the PR chain, or perhaps run HTML-Proofer on demand.

jshaughn:
> So, I wonder if we want to keep it as part of the PR chain, or perhaps run HTML-Proofer on demand.

I agree it's annoying but if we remove it from the automation Im sure we'll start missing some of the real issues.  It doesn't block merge so for me, it's OK.

israel-hdez:
> Note that HTML-Proofer errors are expected

I see some messages saying "internally linking to {...} which does not exist". Those messages are true positives. They should be fixed.

lucasponce:
> I see some messages saying "internally linking to {...} which does not exist". Those messages are true positives. They should be fixed.

+1

lucasponce:
I fixed the "internal errors" from HTTP Proof.
To me this PR is ready from my side.

cc @jshaughn I'd suggest to review it / consolidate it before we have more releases or the work in staging/latest will need to be re-arranged too.

jshaughn (approved):
For me it is good.   I would still like to see more space between About and Features on the landing page but it is not a blocker.  Better to get this merged before today's build.

lucasponce:
Thanks ! it's a big change and it would be good to consolidate.
We can adjust / fix anything after this.

PR #419, opened 2021-10-01
Kiali landing page with Docsy

israel-hdez:
Probably, a little better:

localhost_1313_


israel-hdez:
@jshaughn do we want to keep the old figures for the features?
It would look as follows:

image

jshaughn:
> do we want to keep the old figures for the features?

Honestly, I don't have a preference.  What do you think?  Probably we don't want icons and images, just one or the other.

israel-hdez:
> Honestly, I don't have a preference. What do you think? Probably we don't want icons and images, just one or the other.

For proofing, this is how it looks with _only_ images:

image

I think I also don't have a hard preference.

If it makes the decision easier, using images would require customizing that part of the template, while using icons with font-awesome is already supported (That said, I could open a PR to add support for images in Docsy).

jshaughn:
> If it makes the decision easier, using images would require customizing that part of the template

I'd say let's stick as best as we can to what is provided, and avoid customization.   I don't really care if they go away if it means we can just use the template as-is.  If you want feel strongly and want to pursue a customization for Docsy I'm OK with it.  But as I've been saying, this move it to help remove customization and work on content.

israel-hdez:
OK, the landing page is ready.
I removed the blog section, because docsy doesn't support it.

Once we are ready for customizations, we can recover that part with this code:

```
{{< blocks/section >}}
  {{< blog >}}
{{< /blocks/section >}}
```

israel-hdez:
I updated the screenshot in the main comment with the current look.

israel-hdez:
I also removed the "About" page. Most likely, we won't need it.

abonas:
I like the new look. If it makes things easier, then we can use icons and not images. Definitely not both together :)
I recommend increasing a bit the font size on the section explanations ("Quickly identify issues", etc) - it is really small atm.

jshaughn:
@abonas Thanks for taking a look but this is still WIP, probably best to hold comments at this point.  Thanks.

config.toml
```diff
@@ -93,8 +93,8 @@ weight = 1
 section = ["HTML", "print", "RSS"]
 
 [params]
-copyright = "The Kiali Project"
-privacy_policy = "https://policies.google.com/privacy"
+#copyright = "The Kiali Project"
+#privacy_policy = "https://policies.google.com/privacy"
```
jshaughn, line 97: We probably need valid settings for these.  I'm not sure what they are but we should at least have a copyright, I would think.

layouts/shortcodes/blog.html
```diff
@@ -0,0 +1,41 @@
+{{- $maxBlogPosts := 10 }}
```
jshaughn, line 1: I guess this is a custom shortcode you created to support the blogs?

themes/docsy/assets/scss/_kiali-home.scss
```diff
@@ -321,6 +321,8 @@ section#blog-latest div#blog-blocks > a > div p {
   section#blog-latest div#blog-blocks {
```
jshaughn, line 321: Is this scss actually used?

jshaughn (commented):
couple of non-blocking comments/questions

config.toml
israel-hdez: I chose to remove the copyright notice, because apparently, we are unsure who is. That is what I understand from this discussion that a community user opened: https://github.com/kiali/kiali/discussions/4252. No concrete answer was provided, so I prefer to remove the copyright notice until we are certain.

And about the privacy policy, do we have one?

layouts/shortcodes/blog.html
israel-hdez: Yes. Although it is unused. I may remove it if we won't bring back that block.

themes/docsy/assets/scss/_kiali-home.scss
israel-hdez: I don't know. Probably not.
I noticed it was used in the blog cards. But since I removed it, I'm not sure if it is being used somewhere else. Surely, it is not used in the landing page.

israel-hdez:
@abonas 

> I recommend increasing a bit the font size on the section explanations ("Quickly identify issues", etc) - it is really small atm.

Sometimes I think it's an optical illusion.
The text size on the blue block (that says "Kiali is a management console for Istio...") is a little big IMO. And that may give the illusion that the rest of the text is small.

config.toml
jshaughn: OK, see the e-mail with Josh, I guess we don't really have a single copyright holder.  I'm fine commentiing it out, although I think you could still The Kiali Project, as a representation of all the contributors.

As for the Privacy Policy, I don't know, it could benice to have one but we can keep it commented out.  We could set it https://www.redhat.com/en/about/privacy-policy whether it is commented out, or not.

jshaughn (approved):
I'm fine for merging this whenever you feel like it, the whole branch is a WIP.

PR #421, opened 2021-10-07
Restore blog section

israel-hdez:


israel-hdez:
@jshaughn So, I have two proposals. Yesterday I was checking how to do a dedicated page for the latest blog posts. It's look is as follows:

blog_section

It looks OK, although I'm not sure if both blogs and videos would fit in the same section.

The other proposal is in the landing page as cards:

blog_landing

In the landing page, I'm limiting to only the 3 latest posts.

Personally, after thinking about this a lot, I'm more comfortable with the cards in the landing page. 
Although I like a little more the dedicated page, since we are using Medium, I don't find a good reason to mirror the preview of blogs in the website. I think we simply want that people is aware that we have a blog.

jshaughn:
I think the Blog page looks pretty good and, if it were to have two child pages, one for blogs and one for videos, it could work fine.  But the landing page has the advantage of getting it out in front, and the horizontal cards make scanning things easy.  And, as you say, they can just jump to Medium.  Also, it's what people are used to, and they may like it better if the new site is more similar.  If we can also do the cards for the YouTube videos I think we're good.  I would also recommend adding back a button at the top, to save people from having to scroll down, and also to advertise.  One more advantage, it removes an entry from the top menu, simplifying it a bit.

israel-hdez:
@jshaughn Can you take a look?
I restored the latest videos card.

Also, I removed the dedicated blog page to only keep the cards.

jshaughn:
@israel-hdez This looks really good to me.  Let's promote this PR, do a final review, and merge it in.   One comment, I think we should remove the bottom section (the three badges, not the footer, the footer is good).  I don't think it adds very much, in fact it's sort of redundant and it takes up more space.

israel-hdez:
@jshaughn I agree.
I removed the three badges at the bottom.

I promoted this PR.

jshaughn (approved):
Looks great, @israel-hdez.  I admit to not knowing exactly how things work, but assuming all of the files in the PR are needed, it looks good to me.  In particular, how is layouts/blog/list.html used?

israel-hdez:
@jshaughn Ouch! The `layouts/blog/list.html` is a left over.
I will merge this PR as is, and immediately push a commit directly to the `docsy` branch to remove it.

israel-hdez:
> I will merge this PR as is, and immediately push a commit directly to the `docsy` branch to remove it.

Done!

PR #423, opened 2021-10-11
Add CSS to limit image size, and position slightly offset from

jshaughn:
text (and centered for max-width images)


israel-hdez (approved):
LG!
I can see it fixes the overflow issue.

PR #427, opened 2021-10-13
Cleanup

jshaughn:
- for consistency and simplicity, migrate from html anchors to markdown image syntax
- add some top and bottom margin to img styling so that we can remove a bunch of inline `<br />`
- update remaining `</br>` to recommended `<br />`

...and a couple more change:

- update Releases to currently be master and pre-1.47.
- remove remaining .adoc and update README
-

PR #429, opened 2021-10-13
Add netlify configuration to build the Docsy website and other fixes

israel-hdez:
Also:

* Fix landing page IMG styles
* Set final Google's CSE id.
* Set Google Analytics ID
* Fix privacy policy URL. Although, keep it commented out.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -64,7 +64,7 @@ anchor = "smart"
 [services]
 [services.googleAnalytics]
 # Comment out the next line to disable GA tracking. Also disables the feature described in [params.ui.feedback].
-id = "UA-00000000-0"
+id = "UA-130444728-1"
 
 # Language configuration
 
@@ -94,7 +94,7 @@ section = ["HTML", "print", "RSS"]
 
 [params]
 #copyright = "The Kiali Project"
-#privacy_policy = "https://policies.google.com/privacy"
+#privacy_policy = "https://www.redhat.com/en/about/privacy-policy"
 
 # First one is picked as the Twitter card image if not set on page.
 # images = ["images/project-illustration.png"]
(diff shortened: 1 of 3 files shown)


israel-hdez:
@jmazzitelli @jshaughn OK, once we remove the adoc files, this `netlify.toml` file allows Netlify to build the website.
No need for a new `make` target.

We can merge once #427 goes in.

config.toml
```diff
@@ -64,7 +64,7 @@ anchor = "smart"
 [services]
 [services.googleAnalytics]
 # Comment out the next line to disable GA tracking. Also disables the feature described in [params.ui.feedback].
-id = "UA-00000000-0"
+id = "UA-130444728-1"
```
jmazzitelli, line 67: do we have a Google Analytics account that we are using??

netlify.toml
```diff
@@ -0,0 +1,6 @@
+[build]
+  command = "hugo"
+  publish = "public"
+
+[build.environment]
+  HUGO_VERSION = "0.75.0"
```
jmazzitelli, line 6: we should put a comment in the Makefile where HUGO_VERSION is defined - we should indicate that both the Makefile and this netlify.toml setting should match.

jmazzitelli (approved):
We have netlify previews again!
This LGTM. Unless you want to put a comment in the Makefile as I describe in my earlier comment (`we should put a comment in the Makefile where HUGO_VERSION is defined - we should indicate that both the Makefile and this netlify.toml setting should match.`)

config.toml
jshaughn: It would be good to add a comment here indicating where this value comes from.  And in the same vein, where the search ID comes from as well.

jshaughn (approved):
The CSS fixes look good, blog and video cards are looking good. All good stuff.  Approving, can merge now or after you add some recommended comments.

config.toml
israel-hdez: I'm not sure if it is an "account" but we do have Google Analytics (for quite a long time). I have access to that dashboard.

And about the Google CSE for the custom search, I don't have access to https://cse.google.com/. Most likely, @josejulio knows who owns that ID and how it was created.

josejulio: I added you as an owner of the cse profile.

PR #431, opened 2021-10-14
remove files I don't think we need - edit package.json

jmazzitelli:


package.json
```diff
@@ -1,21 +1,21 @@
...
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/google/docsy-example.git"
+    "url": "git+https://github.com/kiali/kiali.io.git"
   },
-  "author": "",
-  "license": "ISC",
+  "author": "Kiali",
+  "license": "APL 2.0",
```
israel-hdez: Probably, change to "Apache-2.0". That's the short identifier (reference https://www.apache.org/licenses/LICENSE-2.0)

jmazzitelli: done

PR #434, opened 2021-10-14
script to do the release.

jmazzitelli:
be careful when testing. You must set the --remote-name (or just edit the script locally) so it matches your fork name - do not use `origin`.

REPEAT! DO NOT USE `origin` 

When testing, you might even want to remove or rename your `origin` remote so you don't accidentally push up changes when testing this script. Test against your fork.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -7,8 +7,9 @@ if [ -z "${1:-}" ]; then
   exit 1
 fi
 
-# Because the version is used as part of the base image, the version should have dots converted to dashes
-CURRENT_VERSION="${1//./-}"
+# Because the version is used as part of the base image, the version for the base image URL should have dots converted to dashes
+CURRENT_VERSION="${1}"
+CURRENT_VERSION_WITH_DASHES="${CURRENT_VERSION//./-}"
 REMOTE_NAME="${REMOTE_NAME:-origin}"
 CURRENT_BRANCH="${CURRENT_BRANCH:-current}"
 STAGING_BRANCH="${STAGING_BRANCH:-staging}"
@@ -20,7 +21,7 @@ echo "===== Create a new version branch named [${CURRENT_VERSION}] based on bran
 git checkout -b ${CURRENT_VERSION} ${REMOTE_NAME}/${CURRENT_BRANCH}
 
 echo "===== Setting the base URL for the branch [${CURRENT_VERSION}]"
-sed -i "s/baseURL = .*/baseURL = \"https:\/\/${CURRENT_VERSION}.kiali.io\"/" config.toml
+sed -i "s/baseURL = .*/baseURL = \"https:\/\/${CURRENT_VERSION_WITH_DASHES}.kiali.io\"/" config.toml
 git commit -am "Set base URL for branch: ${CURRENT_VERSION}"
 
 echo "===== Push the new version branch [${CURRENT_VERSION}] to remote [${REMOTE_NAME}]"
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -3,11 +3,11 @@
 set -eu
 
 if [ -z "${1:-}" ]; then
-  echo "ERROR! Pass in the new version string"
+  echo "ERROR! Pass in a version string"
   exit 1
 fi
 
-NEW_VERSION="${1}"
+CURRENT_VERSION="${1}"
 REMOTE_NAME="${REMOTE_NAME:-origin}"
 CURRENT_BRANCH="${CURRENT_BRANCH:-current}"
 STAGING_BRANCH="${STAGING_BRANCH:-staging}"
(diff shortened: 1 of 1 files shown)


nrfox:
@jmazzitelli updating a setting in config.toml fixed the right hand nav links: https://github.com/kiali/kiali.io/pull/435. I'm not sure if that will affect the release process but it doesn't seem like it should?

scripts/release.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+set -eu
+
+if [ -z "${1:-}" ]; then
+  echo "ERROR! Pass in the new version string"
+  exit 1
+fi
+
+NEW_VERSION="${1}"
```
jshaughn: It's maybe a little confusing calling this NEW_VERSION.   If we are releasing v1.50 of Kiali, then the content in `staging` is for `v1.50`, the content in `current` is for `v1.49`.   I think perhaps this should be called CURRENT_VERSION.  Does that make any sense?  In the end we want a `v1.49` branch with `v1.49` content, and the `current` and `staging` branches both with `v1.50` content.

jmazzitelli: Yeah, I actually questioned that myself. I'll change to "CURRENT_VERSION" since it corresponds to the version of the 'current' branch.

israel-hdez:
What's the format for the CURRENT_VERSION parameter?

If it's `vX.Y` you should replace dots with dashes and use the resulting string for creating branches and setting the baseURL.

jmazzitelli:
> What's the format for the CURRENT_VERSION parameter?
> 
> If it's `vX.Y` you should replace dots with dashes and use the resulting string for creating branches and setting the baseURL.

Good point. I was assuming the branch names have "vX.Y" in them. But yeah, those go in the URL so it makes sense that dots are converted to dashes. I'll update the script.

jmazzitelli:
OK, I checked in two commits - I'm not sure which one we want to keep.

1. This [commit ce77ce](https://github.com/kiali/kiali.io/pull/434/commits/ce77ce2cd807399c9bf297a9e07dd1096e3b4128) will have BOTH the branch name AND the baseURL have dots converted to dashes in the version string. So the branch name AND baseURL will have something like `v1-36-0` in it.
2. This [commit 4136a2](https://github.com/kiali/kiali.io/pull/434/commits/4136a20b6a10c66405791d3fe91e2c43901aad1f) will allow the branch name to have normal semver strings as is typical for all our other kiali repos (e.g. `v1.36.0`) but the base URL will have dots converted to dashes (`v1-36-0`).

I prefer 2 (https://github.com/kiali/kiali.io/pull/434/commits/4136a20b6a10c66405791d3fe91e2c43901aad1f) if possible - I think our branches should be the same semver format as we use across all our repos. I assume docsy/netlify will allow this - I assume docsy/netlify will see branch names like `vX.Y.Z` and will know to convert those names so the dots become dashes when it publishes to a URL such as `https://vX-Y-Z.kiali.io`.

However, if docsy/netlify does not support this (i.e. if the branch name MUST NOT have dots in them) then we can use commit https://github.com/kiali/kiali.io/pull/434/commits/ce77ce2cd807399c9bf297a9e07dd1096e3b4128 -- that just means all our branches will have names like `vX-Y-Z`.

UPDATE: I didn't even think about our current `pre-v1.41` branch - which HAS dots and which we know works. So I'll go with what I wanted to do (number 2 above).

jmazzitelli:
don't merge this yet until we decide what we want wrt branch names.
I also want to change the way the default remote name is - I don't like the default being "origin" - I see bad things in the future if we keep it like that.

jmazzitelli:
OK. This is ready. I did some testing in my own fork and the branches look correct after the script runs.

You can test like this:

First, copy this new release.sh script in /tmp (because when you run you are checking out branches and this script will go away since it isn't anywhere yet except this PR) and set its execute bit so you can run it:

```
cd /tmp
wget https://raw.githubusercontent.com/kiali/kiali.io/4de6e0700dedd7e77edb7fd8002fe5a89f4a7032/scripts/release.sh
chmod +x /tmp/release.sh
```

Then git clone your own fork of this kiali.io and cd to its directory... because I'm paranoid, do NOT use `origin` as the name of the remote (use -o option and set it to `myfork` for example) and make SURE you are using your own github user's fork and not the remote "kiali" remote... e.g.:

```
git clone -o myfork git@github.com:<your-github-username>/kiali.io.git
cd kiali.io
```

Now run the release, using your "myfork" for the remote name:

```
/tmp/release.sh --current-version v1.99.0 --remote-name myfork
```

You should have pushed to your fork a new branch v1.99 and a new current branch that is based on the new staging branch. To verify:

* v1.99 branch should have the last commit showing it updated base URL of config.toml to use v1-99.
```diff
$ git log -u -1 myfork/v1.99
commit 029262edfbdfcc6d299ce059971d03d8dff3b3c4 (myfork/v1.99, v1.99)
Author: John Mazzitelli <mazz@redhat.com>
Date:   Fri Oct 15 14:06:23 2021 -0400

    Set base URL for branch: v1.99

diff --git a/config.toml b/config.toml
index ad2fd61..564dae3 100644
--- a/config.toml
+++ b/config.toml
@@ -1,4 +1,4 @@
-baseURL = "https://kiali.io"
+baseURL = "https://v1-99.kiali.io"
 title = "Kiali"
```
* `gitk myfork/v1.99` should show it is the same as the previous current (just look at the commits to compare)
* current branch should have the last commit showing it updated the base URL of config.toml to use just kiali.io.
```diff
$ git log -u -1 myfork/current
commit 3172b8894d3b747f004d49e508d916dff7f20bcb (HEAD -> current, myfork/current)
Author: John Mazzitelli <mazz@redhat.com>
Date:   Fri Oct 15 14:06:25 2021 -0400

    Set base URL for branch: current

diff --git a/config.toml b/config.toml
index 0388939..1d5a7c2 100644
--- a/config.toml
+++ b/config.toml
@@ -1,4 +1,4 @@
-baseURL = "https://staging.kiali.io"
+baseURL = "https://kiali.io"
 title = "Kiali"
```
* `git diff myfork/staging..myfork/current` should show current is now the same as staging except for that commit to change base URL:
```diff
$ git diff myfork/staging..myfork/current
diff --git a/config.toml b/config.toml
index df9a1d2..fc397d4 100644
--- a/config.toml
+++ b/config.toml
@@ -1,4 +1,4 @@
-baseURL = "https://staging.kiali.io"
+baseURL = "https://kiali.io"
 title = "Kiali"
```
* staging branch should show the commit that added the new params.versions for 1.99:
 ```diff
$ git log -u -1 myfork/staging
commit 1e844d775840cb5abd9002399c233f0e77ec8687 (myfork/staging, myfork/HEAD, staging)
Author: John Mazzitelli <mazz@redhat.com>
Date:   Fri Oct 15 14:06:24 2021 -0400

    Add params.versions: v1.99

diff --git a/config.toml b/config.toml
index 06ee6c4..0388939 100644
--- a/config.toml
+++ b/config.toml
@@ -222,6 +222,10 @@ enable = false
 
 ## NEXT PARAMS.VERSIONS PLACEHOLDER
 
+[[ params.versions ]]
+  version = "v1.99"
+  url = "https://v1-99.kiali.io"
+
 [[params.versions]]
```

File: scripts/release.sh [hunk 2]
```diff
@@ -0,0 +1,131 @@
...
+
+echo "===== Set baseURL in config.toml for the branch [${CURRENT_VERSION}]"
+sed -i "s/baseURL = .*/baseURL = \"https:\/\/${CURRENT_VERSION_WITH_DASHES}.kiali.io\"/" config.toml
+git commit -am "Set base URL for branch: ${CURRENT_VERSION}"
+
+echo "===== Push the new version branch [${CURRENT_VERSION}] to remote [${REMOTE_NAME}]"
+git push ${REMOTE_NAME} ${CURRENT_VERSION}
+
+echo "===== Create a new branch named [${STAGING_BRANCH}] (or switch to it if it already exists)"
+git checkout -b ${STAGING_BRANCH} ${REMOTE_NAME}/${STAGING_BRANCH} || git checkout ${STAGING_BRANCH}
```
israel-hdez, line 109: Post-merge comment: Can you add a reset if it fallback to the right of the `||`?

We want to be sure that the existent branch is on par with the remote.

(on the following steps where you checkout `CURRENT_BRANCH`, it doesn't matter because you end doing a hard reset to replace everything)

jmazzitelli: good catch - fixed here: https://github.com/kiali/kiali.io/pull/442

PR #435, opened 2021-10-14
Update main github branch and add shell to make

nrfox:
Fixes github links on right hand side of nav to point to "staging" branch instead of the default "master".


jmazzitelli (approved):
this makes sense to me -- cc @jshaughn

jshaughn (approved):
Yep, this looks correct.

PR #443, opened 2021-10-18
Add and fix contribution guide

### LINKED ISSUE #4428: [website] Contribution guidelines link leads to 404
https://kiali.io/docs/contribution-guidelines/

aljesusg:
Fix https://github.com/kiali/kiali/issues/4428

### DIFF
--- a/content/en/community/contribution.md
+++ b/content/en/community/contribution.md
@@ -1,0 +1,59 @@
+---
+title: "Contribute"
+date: 2018-10-31T16:04:38+02:00
+draft: false
+type: docs
+weight: 2
+---
+
+## Kiali is Open Source
+
+Kiali is developed in the open under an [Apache Software License version 2.0](https://www.apache.org/licenses/LICENSE-2.0.txt).
+As such Kiali welcomes contributions in every form.
+This can be code, but also bug reports, feature requests, translations and much more.
+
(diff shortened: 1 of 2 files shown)


israel-hdez:
I'm not sure about this. We are not supposed to change files under `theme/docsy`.

aljesusg:
If we add this under /docs this will appear in the menu of "docs". I think that we don't want to have contribution guide there, right?

israel-hdez:
> If we add this under /docs this will appear in the menu of "docs".

Yes

> I think that we don't want to have contribution guide there, right?

I'm not sure. I was told that the template should guide the structure. This is why I'm saying that, in general, changes to anything inside the `themes` folder should be rejected.

Anyway, I'll let @jshaughn to answer if it's OK to have the contribution guide under `docs`. If not, the template will need to be changed.

jmazzitelli:
> This is why I'm saying that, in general, changes to anything inside the `themes` folder should be rejected.

Yes, in general, we need to just use the theme as it is. The whole point of why we moved to this new doc structure was so we did not have to do any customization like we had to do with the other site.  Just work with the theme we have -- put the docs as best we can in the structure we have in the theme.

aljesusg:
Closing by https://github.com/kiali/kiali/issues/4428#issuecomment-946175334

PR #462, opened 2021-10-27
do not require npm locally

jmazzitelli:
We must do the npm prune inside the container but when /site is mounted to the container. Can't be done in the Dockerfile at build time.

To test this, remove `public/`, `resources/`, and especially `node_modules/` directories (there are in our `.gitignore`, they are not checked in and will be auto-generated). Then remove any existing image you already have, and run the `validate-site` target and it should now work.

```
rm -rf node_modules/ public/ resources/
podman rmi kiali/hugo:latest
make validate-site
```

Before it would only work if you ran `npm prune` locally - which created the `node_modules/` - but this requires the user to have npm installed and in PATH which we do not want to require.  Everything should run inside the container and via the make targets.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,6 @@ RUN tar xzvf /hugo.tar.gz -C / \
 
 # Stuff for docsy
 RUN npm -g -D install postcss postcss-cli autoprefixer
-RUN npm prune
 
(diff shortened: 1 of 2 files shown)


jshaughn (approved):
OK cool, this looks good.

PR #467, opened 2021-11-03
clean up kiali.io html proof checks

jmazzitelli:


Makefile
```diff
@@ -34,11 +39,17 @@ serve: build-hugo
 #     is unsafe because files are moved, renamed, etc.
 # 4. URLs to a folder in repository in a branch; i.e. of the form https://github.com/kiali/kiali/blob/v1.24/whatever
 #   - Same reasoning as previous point.
+# 5. URLs to kiali.io and kiali to edit doc files or create new doc files or create new issues
+# 5. URLs to kiali.io commits
```
jshaughn: 6

jshaughn (approved):
OK, I agree that `master` branch is a bad idea for links.  New `clean` target seems reasonable although did you mean to use it as a dep anywhere?  One comment comment, non-blocking.  Conditionally approving, changes are at your option.

jmazzitelli:
> New `clean` target seems reasonable although did you mean to use it as a dep anywhere?

No. I just wanted a way to clean out the auto-generated files so a person can "start from scratch" when testing things or wants to start "clean".

Makefile
jmazzitelli: pushed a fix for that "6"

PR #475, opened 2021-11-11
Small fixes to "deployment options" and prom/jaeger/grafana pages

israel-hdez:
Related to kiali/kiali#3858


israel-hdez:
@jshaughn Here are the fixes.

The only comment I didn't fix is about mentioning Thanos in the Prometheus-like header.

My rationale is that Thanos may be the most common Prometheus replacement, but there may be others. [We have a user that mentioned Victoria Metrics](https://github.com/kiali/kiali/issues/3631). In [VictoriaMetrics GitHub Readme](https://github.com/VictoriaMetrics/VictoriaMetrics#prometheus-querying-api-usage) it is not mentioned that they implement the `/api/v1/status/config` API endpoint that raised the issue in Thanos. If that's the case, Victoria Metrics users may also be able to use these "thanos" configs.


Also, can we keep the Warning/Note headings on alerts?
I already removed those headings, so I won't push for it too much. It's simply that I like more the heading because it gives better understanding about why some paragraphs are quoted with a color. Also, it seems it is [the pattern suggested in Docsy docs](https://www.docsy.dev/docs/adding-content/shortcodes/#alert).

jshaughn (approved):
Thanks for applying the updates.

> My rationale is that Thanos may be the most common Prometheus replacement, but there may be others. We have a user that mentioned Victoria Metrics. In VictoriaMetrics GitHub Readme it is not mentioned that they implement the /api/v1/status/config API endpoint that raised the issue in Thanos. If that's the case, Victoria Metrics users may also be able to use these "thanos" configs.

OK

> Also, can we keep the Warning/Note headings on alerts?
I already removed those headings, so I won't push for it too much. It's simply that I like more the heading because it gives better understanding about why some paragraphs are quoted with a color. Also, it seems it is the pattern suggested in Docsy docs.

I actually responded to this question in the other PR, but basically, we need to be consistent and during the migration, for readability, I omitted alert titles that reinforce the sidebar color.  To me it's just too verbose.  But, if you want to push for it I'd suggest just getting a vote, and if majority votes for change we'd need to apply it consistently, so in a separate PR, for the whole website.

israel-hdez:
> I actually responded to this question in the other PR, but basically, we need to be consistent and during the migration, for readability, I omitted alert titles that reinforce the sidebar color. To me it's just too verbose. But, if you want to push for it I'd suggest just getting a vote, and if majority votes for change we'd need to apply it consistently, so in a separate PR, for the whole website.

Let's keep as is for now. It's just my personal preference. Perhaps if we see people become lost, we may add the titles.

PR #481, opened 2021-11-16
Docs for audit logs, session options and HTTP options

israel-hdez:
Related to kiali/kiali#3858

### DIFF
--- a/content/en/docs/Configuration/authentication/anonymous.md
+++ b/content/en/docs/Configuration/authentication/anonymous.md
@@ -2,6 +2,7 @@
 title: "Anonymous strategy"
 linktitle: "Anonymous"
 description: "Access Kiali with no authentication."
+weight: 10
 ---
 
 ## Introduction
--- a/content/en/docs/Configuration/authentication/session-configs.md
+++ b/content/en/docs/Configuration/authentication/session-configs.md
@@ -1,0 +1,72 @@
+---
+title: "Session options"
+description: "Session timeout and signing key configuration"
+weight: 60
+---
+
+There are two settings that are available for the users session. The first one
+is the **session expiration time**, which is only applicable to
+[token]({{< relref "token" >}}) and [header]({{< relref "header" >}})
+authentication strategies:
+
+```yaml
+spec:
+  login_token:
+    # By default, users session expires in 24 hours.
+    expiration_seconds: 86400
+```
+
+The session expiration time is the amount of time before the user is asked to
+extend his session by another cycle. It does not matter if the user is actively
+using Kiali, the user will be asked if the session should be extended.
+
+The second available option is the **signing key** configuration which is unset by
+default, meaning that a random 16-characters long signing key will be generated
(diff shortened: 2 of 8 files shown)


content/en/docs/Configuration/authentication/anonymous.md
```diff
@@ -2,6 +2,7 @@
 title: "Anonymous strategy"
 linktitle: "Anonymous"
 description: "Access Kiali with no authentication."
+weight: 10
```
jshaughn, line 5: I had removed these weights and just defaulted to alphabetical, do we need a non-default ordering?  For me it's OK if Session Options is sorted alphabetically, I'm fine leaving them here, though, if you feel it should be at the bottom.

content/en/docs/Configuration/authentication/session-configs.md
```diff
@@ -0,0 +1,72 @@
+---
+title: "Session options"
+description: "Session timeout and signing key configuration"
+weight: 60
+---
+
+There are two settings that are available for the users session. The first one
```
jshaughn:
```suggestion
There are two settings that are available for the user's session. The first one
```

content/en/docs/Configuration/authentication/session-configs.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+The session expiration time is the amount of time before the user is asked to
+extend his session by another cycle. It does not matter if the user is actively
+using Kiali, the user will be asked if the session should be extended.
+
+The second available option is the **signing key** configuration which is unset by
+default, meaning that a random 16-characters long signing key will be generated
+and stored to a secret named `kiali-signing-key` in Kiali's installation
+namespace:
```
jshaughn:
```suggestion
The second available option is the **signing key** configuration, which is unset by
default, meaning that a random 16-character signing key will be generated
and stored to a secret named `kiali-signing-key`, in Kiali's installation
namespace:
```

content/en/docs/Configuration/authentication/session-configs.md
```diff
@@ -0,0 +1,72 @@
...
+CR:
+
+```yaml
+spec:
+  login_token:
+    signing_key: "secret:<secretName>:<secretDataKey>"
+```
+
+{{% alert color="danger" %}}
+It is possible to specify the signing key directly in the Kiali CR in the
```
jshaughn:
```suggestion
It is possible to specify the signing key directly in the Kiali CR, in the
```

jshaughn (approved):
Looks good, just some minor suggestions.  I'll approve and you can apply the changes you like.

content/en/docs/Configuration/authentication/anonymous.md
israel-hdez: It's almost alphabetical. I added weights only to place the session configs at the end.
I find weird that _Session Options_ end in the middle of the auth strategy pages in the navigation menu.

PR #500, opened 2022-01-07
fix: the steps to finish resource creation

Xunzhuo:


jshaughn (commented):
I believe the step numbers can vary depending on what the operator has to do.  So it could be 6, or 7, or something else.  I'll wait for @jmazzitelli to comment but I don't think we need to make a change (unless the change is to explain that there can be a variable number of steps).

jmazzitelli:
> I believe the step numbers can vary depending on what the operator has to do. So it could be 6, or 7, or something else.

Correct. This number can be different depending on what the operator does. Closing.

jmazzitelli:
@Xunzhuo it might be best from now on to ask in the [Kiali Discussions forum](https://github.com/kiali/kiali/discussions) about changes you think need to be made in either code or docs. We can educate you on the different behaviors of Kiali that you might not be familiar with; this will save you time from writing meaningless PRs and will save us time from having to review this churn we are now experiencing with all these little PRs that are getting rejected.

As @lucasponce mentioned earlier, you should see what issues/enhancements might be good for you to work on - Lucas can help with that.

PR #505, opened 2022-01-25
add double quote for authentication.md commend

antigenius0910:
without double quote
```
ᐅ kubectl get secret -n istio-system $(kubectl get sa kiali -n istio-system -o jsonpath={.secrets[0].name}) -o jsonpath={.data.token} | base64 -d  

zsh: no matches found: jsonpath={.secrets[0].name}
```

with double quote
```
 kubectl get secret -n istio-system $(kubectl get sa kiali -n istio-system -o "jsonpath={.secrets[0].name}") -o jsonpath={.data.token} | base64 -d
eyJhbGciOiJSUzI1NiIsI......
```


jmazzitelli (approved):
this is valid for bash as well. So this looks good to support those using zsh.

PR #509, opened 2022-02-04
docs: add OSSM version compatibility with kiali

Xunzhuo:


lucasponce:
Typically product info is added to their sites, but as there may be users that don't know about the relation between Maistra and OSSM and we introduce a few helpers on the releases:
image

I guess it will help to add a note.

jshaughn (approved):
It's fine.

PR #510, opened 2022-02-07
auto-generate the CRD schema docs when the website is released.

jmazzitelli:
This ensures the latest CRD schema documentation is published when we release the new version of the website.


jmazzitelli:
@israel-hdez @jshaughn I put you on this as reviewers to see what you think. I tested the script and it works fine (I tested on my own repo by sending it `--remote-name jmazzitelli`). You will see the Jenkinsfile will call this script immediately before it calls the other "release" script. So this ensures we have the latest auto-generated documentation in staging just prior to copying the staging branch into the new versioned branch. I did not test anything related to Jenkins - which is why @israel-hdez is a reviewer here - to see what he thinks.

jshaughn (approved):
Visual check looks god to me.

One follow-up question, are we ok with a crd doc gen failure stopping the website release?

jmazzitelli:
> One follow-up question, are we ok with a crd doc gen failure stopping the website release?

As I think about it - I think "yes." Because this is now our documentation for the Kiali CR. If the generation fails for some reason, but we keep going, we are going to publish docs that contains potentially (probably?) outdated CRD schema documentation (that is -- the documentation from the LAST version, which potentially (probably?) will be different than the new version).

israel-hdez (approved):
I think it's fine.

But I wonder if it ain't better to  incorporate this into the existing release script?
It would be only about a 2-3 lines addition rather than a whole new script.

jmazzitelli:
> But I wonder if it ain't better to incorporate this into the existing release script?
> It would be only about a 2-3 lines addition rather than a whole new script.

The only reason I did that was because I wasn't sure if we wanted to combine release with auto-gen of docs. In other words, with two separate scripts, we can stop auto-generation of the code by just removing the one line of the Jenkinsfile rather than commenting out code of the release script.

However, if we think it would be better, we can merge the two scripts. But I would make it an option in the release script to generate the docs (so we could turn it off by again just making a one line change to the Jenkinsfile.). Something like `--generate-crd-docs true`.

jmazzitelli:
@israel-hdez @jshaughn see the latest. I merged the auto-gen-doc script into the existing release script. Now the release script can generate docs if we tell it to (which we will tell it to via the Jenkinsfile).

I do the auto-gen of the docs first thing because if that fails, that script will abort and we do not want to abort if we are in the middle of releasing (i.e. we don't want to be in the middle of messing with the current, version, and staging branches - otherwise, it is likely a failure in the auto-gen-doc stuff will cause our branches to be out of whack.)

jshaughn:
Still fine for me.

scripts/release.sh
```diff
@@ -95,6 +101,18 @@ echo "===== SETTINGS"
 echo "===== Fetch the remote content"
 git fetch ${REMOTE_NAME}
 
+if [ "${GENERATE_DOCS}" == "true" ]; then
+  echo "===== Create a new branch named [${STAGING_BRANCH}] (or switch to it if it already exists)"
+  git checkout -b ${STAGING_BRANCH} ${REMOTE_NAME}/${STAGING_BRANCH} || (git checkout ${STAGING_BRANCH} && git reset --hard ${REMOTE_NAME}/${STAGING_BRANCH})
+
+  echo "===== Generate the CRD schema documentation"
+  make gen-crd-doc
+  git commit -am "Auto-generated CRD schema documentation"
```
israel-hdez, line 110: I was expecting seeing only this chunk decorated with the conditional after line 127 to avoid the double checkout and the double push.

But this works, anyways. So if you like this, I'm not against.

israel-hdez (approved):

scripts/release.sh
jmazzitelli: The double checkout is needed - because if we do this in the "middle" and it fails, we would have pushed changes to other branches without finishing the release - thus leaving our branches in a corrupted state. We need to do this first, finish the auto-generation - and only when successful, keep going to the release portion.

jmazzitelli: This is what I mean - if we do the auto-gen after this, we will have pushed the new version branch, but then if auto-gen failed we won't have pushed the staging or current branch and things would get weird at that point.
```
echo "===== Create a new version branch named [${CURRENT_VERSION}] based on branch [${CURRENT_BRANCH}]"
git checkout -b ${CURRENT_VERSION} ${REMOTE_NAME}/${CURRENT_BRANCH}

echo "===== Set baseURL in config.toml for the branch [${CURRENT_VERSION}]"
sed -i "s/baseURL = .*/baseURL = \"https:\/\/${CURRENT_VERSION_WITH_DASHES}.kiali.io\"/" config.toml
git commit -am "Set base URL for branch: ${CURRENT_VERSION}"

echo "===== Push the new version branch [${CURRENT_VERSION}] to remote [${REMOTE_NAME}]"
git push ${REMOTE_NAME} ${CURRENT_VERSION}
```

israel-hdez: Well... there are several pushes here. I think a failure in almost any of them would require a manual fix.
But OK, let's leave this here to be more on the safe side.

PR #515, opened 2022-03-09
fix: Update the ingress spec for the Example CR

saltbo:
The ingress spec is outdate

### DIFF
--- a/content/en/docs/Configuration/kialis.kiali.io.md
+++ b/content/en/docs/Configuration/kialis.kiali.io.md
@@ -155,9 +155,12 @@ spec:
           - http:
               paths:
               - path: "/kiali"
+                pathType: Prefix
                 backend:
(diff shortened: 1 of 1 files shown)


jmazzitelli:
The file you changed `content/en/docs/Configuration/kialis.kiali.io.md` is auto-generated. Do not modify it. The source for this file is found in two places in the operator repo:
* The example CR (which is what I think you want to update): https://github.com/kiali/kiali-operator/blob/master/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml
* The CRD docs: https://github.com/kiali/kiali-operator/blob/master/crd-docs/crd/kiali.io_kialis.yaml

The operator's `validate-cr` make target is used to confirm the example CR is correct.
The file is generated via `gen-crd-docs` target.

Closing since this PR is incorrect and invalid.

PR #530, opened 2022-05-17
Kiali site schedule release

leandroberetta:
This PR adds the scheduling mechanism.

Next release will be on the next Monday after the end of the sprint triggered manually, merge after that date.

Epic: https://github.com/kiali/kiali/issues/4827

### Note 1

According to the following link: https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow:

`Warning: To prevent unnecessary workflow runs, scheduled workflows may be disabled automatically. When a public repository is forked, scheduled workflows are disabled by default. In a public repository, scheduled workflows are automatically disabled when no repository activity has occurred in 60 days.`

So, scheduled workflows in forks are disabled but just in case I added a condition on important jobs (like pushing to Quay) that for the job to run the GitHub repository that the workflow is running needs to match to "kiali/kiali" (in the case of the server).

### Note 2

In the past, we run the release pipeline every Friday thanks to a cron, then the snapshots or the minor release were created. With this pipelines is the same, we need to run the job every Friday (cron does not support running for example every 3 weeks), but as we removed snapshots, there is not much to do on the first two weeks of our sprint, only to skip the release. On the third week of the sprint, he cron will fire on the Friday, and it will create a minor release.

### DIFF
--- a/content/en/news/release-notes.md
+++ b/content/en/news/release-notes.md
@@ -22,6 +22,7 @@ Features:
 * [Investigate update of Patternfly to be compatible with OS Console](https://github.com/kiali/kiali/issues/4836)
 * [Investigate testing demos scripts on upstream + OpenShift platform](https://github.com/kiali/kiali/issues/5045)
 * [Schedule release pipelines](https://github.com/kiali/kiali/issues/5029)
+* [Update to use new HPA v2](https://github.com/kiali/kiali-operator/pull/536)
 
 Fixes:
 


jmazzitelli (approved):
this just looks like the same changes made to the other repo workflows.. lgtm

israel-hdez (approved):
I echo what Mazz said.

PR #537, opened 2022-05-27
created new section, how to set up development environment

josunect:
I've added a new section about how to set up the development environment, under the Contribution guidelines.
I think that might help, this is just some notes taken with the information provided by the team. 

cc @lucasponce

PR #542, opened 2022-06-08
Added script to ignore new/renamed files for html proofer validator

### LINKED ISSUE #5136: kiali.io CI is validating against staging.kiali.io
HTML-Proofer is validating the url staging.kiali.io instead of the netlify link that the PR deployment is creating. 

See: https://github.com/kiali/kiali.io/runs/6687054315?check_suite_focus=true

`External link https://staging.kiali.io/docs/contribution-guidelines/development-environment/ failed: 404 No error`

Instead of checking:

`https://deploy-preview-537--kiali.netlify.app/docs/contribution-guidelines/development-environment/`

cc @jmazzitelli

josunect:
Added script to ignore new/renamed files for html proofer validator

Fixes [#5136](https://github.com/kiali/kiali/issues/5136)
https://github.com/kiali/kiali.io/pull/537


josunect:
The 404 is not a false positive. It has been removed from Itsio documentation. Not sure if there is any replacement page.

image

Makefile
```diff
@@ -64,6 +64,10 @@ URL_IGNORE=\#$\
           ,/.*web.libera.chat.*/$\
           ,/^http://tracing.istio-system.*/$\
           ,/^https://tracing-service.*/
+
+NEW_URLS=$(shell scripts/ignore_new_urls.sh)
+URL_IGNORE:=$(URL_IGNORE)$(NEW_URLS)
```
jmazzitelli: I am not familiar with this kind of syntax in a makefile. I think you just want URL_IGNORE to be the stdout that is printed by the script right? This seems to be what you want:

```suggestion
URL_IGNORE ?= $(shell scripts/ignore_new_urls.sh 2> /dev/null)
```

Note the `?=` means that the caller of the makefile can override this if they want via the make cmdline option ` -e` (useful for testing the make target, for example).

We also want that `2> /dev/null` because we want to redirect the stderr of the script to dev/null because you don't want error messages considered URLs which might cause errors later in the make invocation.

scripts/ignore_new_urls.sh
```diff
@@ -0,0 +1,26 @@
...
+
+  while IFS= read -r line; do
+  if [[ "$line" =~ .*"$SUB".* ]]; then
+
+              SUBSTR=$(echo $line | sed 's,content/en/docs/,,g')
+              PATTERN=$(echo "${SUBSTR%.*}")
+              URL=$(echo ",/.*${SUBSTR%.*}.*/")
+              URL=$(echo "$URL" | awk '{print tolower($0)}')
+              EXCLUDE_URLS="$EXCLUDE_URLS$URL"
+      fi;
```
jmazzitelli: trivial - but for better readability, if statement should be formatted -- at least line up `fi` with `if` on the same column :) 
```suggestion
  if [[ "$line" =~ .*"$SUB".* ]]; then
    SUBSTR=$(echo $line | sed 's,content/en/docs/,,g')
    PATTERN=$(echo "${SUBSTR%.*}")
    URL=$(echo ",/.*${SUBSTR%.*}.*/")
    URL=$(echo "$URL" | awk '{print tolower($0)}')
    EXCLUDE_URLS="$EXCLUDE_URLS$URL"
  fi;
```

jmazzitelli: Unless there is a reason for doing these separately, I would also consider combining the two statements setting URL - you can just use pipe to combine - you are using pipe anyway on the second line - so it seems you can make this simpler by just combining the two lines:

```
URL=$(echo ",/.*${SUBSTR%.*}.*/" | awk '{print tolower($0)}')
```

jmazzitelli (changes requested):

Makefile
josunect: Agree with `2> /dev/null`

But, in the case of 
`URL_IGNORE ?= $(shell scripts/ignore_new_urls.sh 2> /dev/null)`

I would like to concat the existent defined urls with the ones that the script returns - Should be dynamic to just avoid these 404. The ones already defined in the makefile as URL_IGNORED they always should be ignored - I think, I didn't want to change this behavior.

scripts/ignore_new_urls.sh
josunect: Yes, looks simpler. I will add the change.

Makefile
jmazzitelli: > I would like to concat the existent defined urls with the ones that the script returns

Oh, duh! I didn't notice URL_IGNORE was defined above. I get it now :)

scripts/ignore_new_urls.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+# Used by html-proffer to ignore new/copied/renamed files
```
jmazzitelli: trivial - typo in comment
```suggestion
# Used by html-proofer to ignore new/copied/renamed files
```

jmazzitelli (approved):
this all looks correct (except for that minor spelling typo in the comment). approving.

jmazzitelli:
@josunect rebase this on the latest staging and that CI failure should go away - Jay fixed that bad link in another PR yesterday

jmazzitelli (approved):

josunect:
@jmazzitelli could be possible also to merge it? I cannot do it because I don't have permissions. 

Thanks!

PR #549, opened 2022-06-23
document the operator RELATED_IMAGE_kiali_default setting

### LINKED ISSUE #5238: Document ability to set default Kiali CR image_name in operator from helm charts
We want to add ability to set default Kiali CR image_name in operator using some env variables and use this env variable in helm chart while deploying. Currently, default image_version which operator tries is "quay.io/kiali/kiali" (which is hardcoded), we want to parameterise it so that we can configure(point) it to our container registry during operator deployment.

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/5238

netlify link: https://deploy-preview-549--kiali.netlify.app/docs/installation/installation-guide/advanced-install-options/#change-the-default-image

### DIFF
--- a/content/en/docs/Installation/installation-guide/advanced-install-options.md
+++ b/content/en/docs/Installation/installation-guide/advanced-install-options.md
@@ -73,6 +73,41 @@ spec:
     image_version: your_custom_tag
 ```
 
+### Change the default image
+
+As explained earlier, when you install the Kiali Operator, it will be
+configured to install a Kiali Server whose image will be pulled from quay.io
+and whose version will be the same as the operator. You can ask the operator to
+use a different image by setting `spec.deployment.image_name` and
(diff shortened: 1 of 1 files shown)


jmazzitelli:
FWIW: I did just confirm this feature works.

Run the helm install command that are in these new docs, then create a Kiali CR (withOUT specifying any spec.deployment.image_name or image_version) and you will see the Kiali Server will attempt to pull the image specified in the RELATED_IMAGE_kiali_default variable.

jmazzitelli:
Also, the CI error is not related - I created another github issue for us to fix that - its a 404 for an istio doc page that has been deleted. See: https://github.com/kiali/kiali/issues/5244

leandroberetta (approved):
Verified using an image from my Quay repo:

```
Containers:
  kiali:
    Image:         quay.io/leandroberetta/kiali:v1.50
```

PR #550, opened 2022-06-29
Added note about decreasing the severity of KIA0209 in case of valid subset label.

hhovsepy:
Related to https://github.com/kiali/kiali/pull/5253


jshaughn (approved):
this should also be applied to staging branch, I will create a PR

PR #556, opened 2022-07-12
Warning severity on KIA0104, KIA0202, KIA1101 when ALLOW_ANY

hhovsepy:
When meshConfig.outboundTrafficPolicy.mode is set ALLOW_ANY (or default), put Warning severity on KIA0104, KIA0202, KIA1101.

RFE https://github.com/kiali/kiali/issues/5235

### DIFF
--- a/content/en/docs/Features/validations.md
+++ b/content/en/docs/Features/validations.md
@@ -75,8 +75,8 @@ Either remove the host from the list, correct if there is any typo or deploy a n
 
 #### Severity
 
-<i class="fas fa-times-circle"></i> Error - if OutboundTrafficPolicy.Mode is set to REGISTRY_ONLY
-<i class="fas fa-exclamation-triangle"></i> Warning - if OutboundTrafficPolicy.Mode is set to ALLOW_ANY (default)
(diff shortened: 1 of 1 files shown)


israel-hdez (approved):
This is OK, although I think it will be much, much clearer if this is commented under the `Severity` header.

I don't see wrong with writing bullets under the `Severity` header. Something like:

```
#### Severity

- <i class="fas fa-times-circle"></i> Error - if OutboundTrafficPolicy.Mode is set to REGISTRY_ONLY
- <i class="fas fa-exclamation-triangle"></i> Warning - if OutboundTrafficPolicy.Mode is set to ALLOW_ANY
```

I'm approving anyway, because I'm also OK with this. So, it is up to you to do the change, if you like it.

hhovsepy:
@israel-hdez good idea, have changed.

israel-hdez:
@hhovsepy I think you missed the dash at the beginning of each line so that it shows as bullets. Right now it is on a single line:

image

Unless this is the format you want, perhaps do the fix?

hhovsepy:
@israel-hdez thanks for catch, have fixed.

PR #558, opened 2022-07-19
Update the relnotes script to generate Release notes

josunect:
I was running this script using python 3.8.13 and I get information empty from the project information: 

```
scripts/relnotes.py v1.53 59 TOKEN
Traceback (most recent call last):
  File "scripts/relnotes.py", line 77, in <module>
    projectName = project["name"]
TypeError: 'NoneType' object is not subscriptable
```
The result of the query was: 

`{'data': {'organization': {'name': 'Kiali - Service Mesh Management for Istio', 'project': None}}}`

It looks like it was updated to use the object ProjectV2 (https://docs.github.com/en/issues/trying-out-the-new-projects-experience/using-the-api-to-manage-projects)

I have updated the script using ProjectV2 object, it gets the project information with the same parameter. 

Note: It retrieves the Issue items (Not the Pull requests) and it filters by the state CLOSED and MERGED (Even this is just for Pull Requests, I think).

### DIFF
--- a/scripts/relnotes.py
+++ b/scripts/relnotes.py
@@ -56,16 +56,6 @@ query = """
                                }
                             }
                        }
-                       ...on PullRequest {
-                           title
-                           url
-                           state
-                           labels(first:10) {
-                                 nodes{
-                                       name
-                                 }
-                           }
-                       }
                      }
                 }
             }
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
Thanks so much for updating this, @josunect!  I feel I should apologize, this script was sort of a hack, I never wrote any python or graphql before, or after :-)   Nice job figuring out the graphql changes for github's v2.   Worked for me, I see the same 1.54 changes as you already submitted to staging.

PR #559, opened 2022-07-19
Use html proofer version 3.19.4

josunect:
This PR just specifies the html-proofer version in the Dockerfile. It was upgraded to a major version (from 3 to 4), and
the compatibility is broken. While they are not addressed (https://github.com/kiali/kiali/issues/5313), this forces
to use the version 3.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ RUN tar xzvf /hugo.tar.gz -C / \
 RUN npm -g -D install postcss postcss-cli autoprefixer
 
 # Install html-proofer
-RUN gem install html-proofer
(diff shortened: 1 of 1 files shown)


jmazzitelli (approved):
@josunect do we want this merged or do we want to go to version 4? I'm ok if we just merge this if it fixes the problem

josunect:
> @josunect do we want this merged or do we want to go to version 4? I'm ok if we just merge this if it fixes the problem

I think we can merge this as a workaround to have the validation working while we don't fix this: https://github.com/kiali/kiali/issues/5313

I will add to revert this change when https://github.com/kiali/kiali/issues/5313 is done.

PR #563, opened 2022-07-25
Avoid commiting crd docs if there is nothing to commit

leandroberetta:
This fix will check if there are changes to commit handling the exit code form the git commit command.


jmazzitelli (approved):
LGTM -- did a quick smoke test of this new code snippet and it appears to do what we want it to do

PR #565, opened 2022-08-04
Prometheus tuning and Metric Thinning

jshaughn:
Part-of https://github.com/kiali/kiali/issues/5151

Restructure p8s-jaeger-grafana config with dedicated pages
- Add new content about prom tuning / metric thinning

Relevant links:
Previous page now a header into dedicated pages: 
https://deploy-preview-565--kiali.netlify.app/docs/configuration/p8s-jaeger-grafana/
https://deploy-preview-565--kiali.netlify.app/docs/configuration/p8s-jaeger-grafana/grafana/
https://deploy-preview-565--kiali.netlify.app/docs/configuration/p8s-jaeger-grafana/jaeger/
https://deploy-preview-565--kiali.netlify.app/docs/configuration/p8s-jaeger-grafana/prometheus/

Prometheus page with new content starting at "Prometheus Tuning".  This is the meat of the PR and describes how to configure the potential thinning.
https://deploy-preview-565--kiali.netlify.app/docs/configuration/p8s-jaeger-grafana/prometheus/#prometheus-tuning

PR #566, opened 2022-08-08
Add typhoeus to validation to control timeout and concurrency settings

josunect:
There are many status code 0 errors from the validation (See last validations https://github.com/kiali/kiali.io/pull/565)
With [typhoeus](https://github.com/typhoeus/typhoeus#other-curl-options) it is possible to control parameters like the time out. 

The tool is documented in the html-proofer documentation: 
https://github.com/gjtorikian/html-proofer#readme

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,7 @@ RUN gem install html-proofer
 # Confirm htmlproofer binary is available and show its version
 RUN htmlproofer --version
 
+# This allows us to increase timeouts, and avoid htmlproofer errors
 RUN gem install typhoeus
 
 # Confirm hugo binary is available and show its version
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -31,5 +31,7 @@ RUN gem install html-proofer
 # Confirm htmlproofer binary is available and show its version
 RUN htmlproofer --version
 
+RUN gem install typhoeus
```
jshaughn, line 35:
```suggestion
# This allows us to increase timeouts, and avoid htmlproofer errors
RUN gem install typhoeus
```

jshaughn (approved):
@josunect , thanks for investigating the htmlproofer issues in other prs.  One small suggested addition (just a comment).

Dockerfile
josunect: sounds good!

PR #568, opened 2022-08-11
Fix regex

jshaughn:
https://deploy-preview-568--kiali.netlify.app/docs/configuration/p8s-jaeger-grafana/prometheus/#metric-thinning-with-crippling


jmazzitelli (approved):
punctuation deficiency has been corrected.

PR #573, opened 2022-08-16
fixed typo in docs

ScriptingShrimp:
found one small typo in docs


josunect (approved):
Tested and changes LGTM. 

Thanks for the PR, @ScriptingShrimp 

Tested the validator locally and it passed. Probably it will need some more adjustments.

PR #581, opened 2022-09-15
be more explicit about why operator is recommended

jmazzitelli:
netlify: https://deploy-preview-581--kiali.netlify.app/docs/installation/installation-guide/install-with-helm/


jshaughn (approved):
The warning may even be better above the introduction, but not a blocker.

PR #584, opened 2022-09-26
trim releases to help ensure the menu doesn't get bottom-cropped

jshaughn:
I had to trim it down such that it gets one entry per ossm version and one entry per unsupported istio version.  Otherwise it's just too long given out 3 week cadence.  As I mentioned, a person could just type in the version they want, it's still out there.


jmazzitelli (approved):
gorgeous
image

PR #589, opened 2022-10-13
Update prometheus.md

chase-childers:
Fix Link to Kiali CR Reference Page


jshaughn (approved):
We have an issue in the test automation, but the change looks good to me. Thanks!

PR #595, opened 2022-11-07
Fix error in the image generation due to incompatibilities in the rub…

### LINKED ISSUE #5610: [kiali.io] Fix Error in CI
Error in CI: 

```
STEP 12/15: RUN gem install html-proofer
ERROR:  Error installing html-proofer:
	io-event requires Ruby version >= 3.0. The current ruby version is 2.7.3.183.
Successfully installed public_suffix-5.0.0
Successfully installed addressable-2.8.1
Successfully installed fiber-local-1.0.0
Successfully installed console-1.16.2
Error: error building at STEP "RUN gem install html-proofer": error while running runtime: exit status 1
make: *** [Makefile:32: build-hugo] Error 125
Error: Process completed with exit code 2.
```

Log example of failed job: 

https://github.com/kiali/kiali.io/actions/runs/3398067570/jobs/5650785179#step:3:464

josunect:
Fix error in the image generation due to incompatibilities in the ruby and html proofer versions.

Ruby Version: 
- The new version (5.1) of htmlproofer requires ruby 3
- Ruby 3.1 is the default in Fedora 36

htmlproofer errors: 
- The new html proofer version changed the way in which we should pass the new arguments. Ex. 
`--check_external_hash false`

Was changed by: 
`-no-check-internal-hash`

Fixes https://github.com/kiali/kiali/issues/5610


josunect:
Still need to investigate why we are getting these errors: 

```
* At ./public/search/index.html:131:

  External link https://v1-58.kiali.io/ failed (status code 404)
```

I didn't got that locally. 

Another issue is: 

```
External link https://coreos.com/operators/ failed with something very wrong.
It's possible libcurl couldn't connect to the server, or perhaps the request timed out.
Sometimes, making too many requests at once also breaks things. (status code 0)
```
But this really doesn't exist (4 occurrences)

jshaughn (commented):
Looks like the initial issue is fixed, thanks.  Now we are back to the issues prior to the build issue.  Let's see if we can make progress there.  If we can't solve it fairly easily then perhaps we should disable the checking.

josunect:
A couple of notes: 

- It was doing a validation for a couple of versions in kiali.io (https://v1-XX.kiali.io) that where not available in kiali.io, so they are ignored in the validation (The current version and the previous). 
- Sometimes there are lots of errors in the validation, I think this is due to some kind of concurrency issue. The thypoeus parameters have been changed. But I don't see it very stable yet, so it may require some more adjustments. 

There are many commits as I didn't have these issues in my local environment.

jshaughn:
> It was doing a validation for a couple of versions in kiali

I'm fine if the CI validates only staging.

Makefile
```diff
@@ -68,7 +68,16 @@ URL_IGNORE=\#$\
...
 
+# Skip validation from (Probably) not released version (last one)
+# Get current version from release notes
+VERSION=$(shell egrep -m1 '\#\#' content/en/news/release-notes.md | sed 's/\#\# //' | sed 's/.0//' | sed 's/\./-/' )
+URL_VERSION=,/^https://v$(VERSION).kiali.io/
+URL_IGNORE:=$(URL_IGNORE)$(URL_VERSION)
+VERSION=$(shell egrep -m2 '\#\#' content/en/news/release-notes.md | sed 's/\#\# //' | sed 's/.0//' | sed 's/\./-/' | tail -n1)
+URL_VERSION=,/^https://v$(VERSION).kiali.io/
+URL_IGNORE:=$(URL_IGNORE)$(URL_VERSION)
+
```
jshaughn: This is fine although would it be easier to use sort of a `not equal 'staging'` logic?

jshaughn (approved):
LG! One comment but you can ignore it if it doesn't make sense.

Makefile
josunect: Good idea, @jshaughn ! I will see how to improve it.

jshaughn (approved):
LG!

PR #612, opened 2022-12-14
[staging] Add deprecation notice to v1.60 release notes

### LINKED ISSUE #4705: Deprecate and remove "implicit" oauth flow
Kiali currently supports both ["implicit"](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitFlowAuth) and ["authorization code"](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth) oauth flows. There are security benefits to using the authorization code flow vs. the implicit flow. In addition, moving to the auth code flow would allow Kiali to take advantage of the widely used [go-oidc](https://github.com/coreos/go-oidc) lib.

Pros:
- Improved security
- Less to maintain
- Less to test

Cons:
- Would break providers who only support the "implicit" flow. (I'm not sure these exist)
- May require additional configuration for existing open id users

israel-hdez:
Fixes kiali/kiali#4705

### DIFF
--- a/content/en/news/release-notes.md
+++ b/content/en/news/release-notes.md
@@ -25,6 +25,10 @@ Fixes:
 * [(kiali.io) Aditional CI issues](https://github.com/kiali/kiali/issues/5631)
 * [Kiali heatmap tooltips are too compute-heavy](https://github.com/kiali/kiali/issues/5634)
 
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
LG!

PR #615, opened 2023-01-05
Added Error message for KIA0401

hhovsepy:
Related to https://github.com/kiali/kiali/pull/5693

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1,4 +1,4 @@
-baseURL = "https://staging.kiali.io"
+baseURL = "https://kiali.io"
 title = "Kiali"
 
 enableRobotsTXT = true
--- a/content/en/docs/Features/validations.md
+++ b/content/en/docs/Features/validations.md
@@ -891,6 +891,32 @@ Add missing Service Entry which address will match the Workload Entry's address.
 - [Validator source code](https://github.com/kiali/kiali/tree/v1.52.0/business/checkers/serviceentries/workload_entry_address_match.go)
 
 
+## K8s HTTPRoutes {#httproutes}
+
+### KIA1401 - HTTPRoute is pointing to a non-existent K8s gateway
+
+Gateway API HTTPRoute could be pointing to a [k8s] Gateway that the Route wants to be attached to. When the namespace field is not specified it takes Gateways from the current HTTPRoute's namespace. Here the error indicates that the referenced Gateway is not found in the provided namespace.
+
+#### Resolution
+
+Fix the parentRefs field to target to an existing gateway.
+
+#### Severity
+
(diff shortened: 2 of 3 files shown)


config.toml
```diff
@@ -1,4 +1,4 @@
-baseURL = "https://staging.kiali.io"
+baseURL = "https://kiali.io"
```
jshaughn, line 1: hmm, I guess this is correct, not sure why it was set to staging.kiali.io.

content/en/docs/Features/validations.md
```diff
@@ -891,6 +891,32 @@ Add missing Service Entry which address will match the Workload Entry's address.
 - [Validator source code](https://github.com/kiali/kiali/tree/v1.52.0/business/checkers/serviceentries/workload_entry_address_match.go)
 
 
+## K8s HTTPRoutes {#httproutes}
+
+### KIA1401 - HTTPRoute is pointing to a non-existent K8s gateway
+
+Gateway API HTTPRoute could be pointing to a Gateway (K8s) the Route wants to be attached to. When the namespace field it not specified, it takes Gateways from currect HTTPRoute's namespace. Here the error indicates that the pointed Gateway is not found in the provided namespace.
```
jshaughn:
```suggestion
Gateway API HTTPRoute could be pointing to a [k8s] Gateway that the Route wants to be attached to. When the namespace field is not specified it takes Gateways from the current HTTPRoute's namespace. Here the error indicates that the referenced Gateway is not found in the provided namespace.
```

jshaughn (changes requested):

config.toml
hhovsepy: It seems I did a wrong rebase.

jshaughn (approved):
LG!  Does this also need to go into Staging?  If only merged to Current it will not survive after the next release..

hhovsepy:
Then isn't it better to send into Staging only?

jshaughn:
In general changes go to `staging` because then it will show up on the next release and will live on.  `staging` is basically "master" for kiali.io.   `current` is the most recent release (so in this case it reflects v1.62).  Sometimes we also want to backport to `current` if the change is relevant to code that is already released.  So yes, in general you want `staging`, sometimes also `current`, but very rarely only `current`.  If `KIA0401` is relevant to our released version, v1.62, then yes we want a PR for `current` but  Probably you want to re-target this to `staging`, or have an addition PR for `staging`.

hhovsepy:
Thank you Jay, not it is clear. Closing this PR in favor of https://github.com/kiali/kiali.io/pull/617.

PR #631, opened 2023-02-02
Rework RBAC documentation

### LINKED ISSUE #5720: Rework RBAC documentation
Current RBAC documentation is here: https://github.com/kiali/kiali.io/blob/6311dcf0fe8fe9aa8d39b48ae470d67935a4c607/content/en/docs/Configuration/rbac.md

This documentation was fine before the Kiali cache was introduced. When the Kiali cache was introduced, it became inaccurate as users just need read privileges to namespaces to gain visibility to several other resources through Kiali (i.e. people may just need what we have documented as the _minimum required privileges to login_ and Kiali will let users read everything else).

Probably, this was overlooked because the Kiali cache was opt-in at first. But when the cache transitioned to be turned on by default, the RBAC became "loose by default" and this isn't documented. So, docs are outdated already and are due for an update, although the documented behavior can still be obtained by turning off the Kiali cache.

However, since #5599 is already in progress and people won't be able to turn off the Kiali cache, the documented RBAC capabilities no longer stand and docs need to be updated.

israel-hdez:
Since Kiali is no longer allowing to turn off the kube cache, the RBAC documentation no longer stands. It was anyway due an update.

This is a re-write of the RBAC documentation to reflect current behavior.

Fixes kiali/kiali#5720

### DIFF
--- a/content/en/docs/Configuration/authentication/openid.md
+++ b/content/en/docs/Configuration/authentication/openid.md
@@ -15,7 +15,7 @@ third-party system.
 If your
 [Kubernetes cluster is also integrated with your OpenId provider](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens),
 then Kiali's `openid` strategy can offer 
-[namespace access control]({{< relref "../rbac" >}}) for more details.
+[namespace access control]({{< relref "../rbac" >}}).
 
 Kiali only supports the _authorization code flow_ of the [OpenId Connect spec](https://openid.net/connect/).
 
--- a/content/en/docs/Configuration/authentication/openid.md
+++ b/content/en/docs/Configuration/authentication/openid.md
@@ -14,9 +14,8 @@ third-party system.
 
 If your
 [Kubernetes cluster is also integrated with your OpenId provider](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens),
-then Kiali's `openid` strategy can offer role-based access control (RBAC) through the
-[Kubernetes authorization mechanisms](https://kubernetes.io/docs/reference/access-authn-authz/rbac/). See the
-[RBAC documentation]({{< relref "../rbac" >}}) for more details.
+then Kiali's `openid` strategy can offer 
+[namespace access control]({{< relref "../rbac" >}}) for more details.
 
 Kiali only supports the _authorization code flow_ of the [OpenId Connect spec](https://openid.net/connect/).
 
@@ -26,10 +25,10 @@ The [Kiali's signing key]({{< relref "session-configs" >}}) needs to be 16, 24
 or 32 byte long. If you install Kiali via the operator and don't set a custom
 signing key, the operator should create a 16 byte long signing key.
 
-If you *_don't need_* RBAC support, you can use any
+If you *_don't need_* namespace access control support, you can use any
 working OpenId Server where Kiali can be configured as a client application.
 
-If you *_do need_* RBAC support, you need either:
+If you *_do need_* namespace access control support, you need either:
 
 * A [Kubernetes cluster configured with OpenID connect integration](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens),
 which results in the API server accepting tokens issued by your identity provider.
@@ -48,7 +47,7 @@ strategy by using
 reverse proxy that handles the OpenID authentication and forwards the
 authenticated requests to the Kubernetes API.
 
-## Set-up with RBAC support {#setup-with-rbac}
+## Set-up with namespace access control support {#setup-with-rbac}
 
 Assuming you already have a working Kubernetes cluster with OpenId integration
 (or a working alternative like `kube-oidc-proxy`), you should already had
@@ -114,7 +113,7 @@ attribute is the URI of the reverse proxy or cluster API replacement (only
 HTTPS is allowed). The `api_proxy_ca_data` is the public certificate authority
 file encoded in a base64 string, to trust the secure connection.
 
-## Set-up with no RBAC support
+## Set-up with no namespace access control support
 
 Register Kiali as a client application in your OpenId Server. Use the root path
 of your Kiali instance as the callback URL. If the OpenId Server provides you a
@@ -144,7 +143,7 @@ spec:
 ```
 
 {{% alert color="warning" %}}
-As RBAC is disabled, all users logging into Kiali
+As namespace access control is disabled, all users logging into Kiali
 will share the same cluster-wide privileges.
 {{% /alert %}}
 
@@ -164,10 +163,11 @@ spec:
       username_claim: "email"
 ```
 
-If you enabled RBAC, you will want the `username_claim` attribute to match the
-`--oidc-username-claim` flag used to start the Kubernetes API server, or the
-equivalent option if you are using a replacement or reverse proxy of the API
-server. Else, any user-friendly claim will be OK as it is purely informational.
+If you enabled namespace access control, you will want the `username_claim`
+attribute to match the `--oidc-username-claim` flag used to start the
+Kubernetes API server, or the equivalent option if you are using a replacement
+or reverse proxy of the API server. Else, any user-friendly claim will be OK as
+it is purely informational.
 
 ### Configuring requested scopes {#configure-scopes}
 
@@ -406,7 +406,7 @@ refresh your Kiali Pod to _set_ the Secret if you add the Secret after the Kiali
 {{% alert color="warning" %}}
 The OpenID authentication strategy can be used
 with Azure Kubernetes Service (AKS) and Azure Active Directory (AAD) with Kiali
-versions 1.33 and later. Prior Kiali versions do not support RBAC on Azure.
+versions 1.33 and later. Prior Kiali versions do not support namespace access control on Azure.
 {{% /alert %}}
 
 AKS has support for a feature named _AKS-managed Azure Active Directory_, which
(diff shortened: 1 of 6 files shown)


content/en/docs/Configuration/rbac.md
```diff
@@ -1,120 +1,138 @@
...
+operations which may be sensitive, the users will need to have the required
+privileges in the cluster to perform updates - i.e. the cluster RBAC takes
+effect.
 
-```
-$ kubectl create rolebinding john-openid-binding --clusterrole=kiali --user="john@example.com" --namespace=mynamespace
-$ oc adm policy add-role-to-user kiali john -n mynamespace # For OpenShift clusters
-```
+{{% alert color="warning" %}}
+Kiali is going to reject login to users that aren't authorized any namespace.
```
jmazzitelli:
```suggestion
Kiali is going to reject login to users that aren't authorized to see any namespace.
```

jmazzitelli (dismissed):

content/en/docs/Configuration/rbac.md
israel-hdez: Fixed.

israel-hdez:
This should be ready for review.

Preview link to the reworked page: https://deploy-preview-631--kiali.netlify.app/docs/configuration/rbac/

jmazzitelli:
Documenting the actual Role and all its permissions could be prone to "rot" (i.e. as we change the permissions in the code base in the future, this documentation would have to change, in fact depending on the version of Kiali you are using, these docs would be different). I guess we could ask the user to just look at the versioned kiali.io docs for the different versions of the Roles, but what if we did this instead? In the docs tell them to generate the ClusterRole from the server helm-chart - this generates a ClusterRole. If they want a per-namespace Role, they can edit the resulting yaml and change the Kind from ClusterRole to Role. For example, to generate a Kiali v1.57 Server ClusterRole with view_only permissions, you can do this:

```
helm template --version 1.57.0 --set deployment.view_only_mode=true --show-only templates/role-viewer.yaml kiali-server kiali/kiali-server
```

This results in:

```yaml
---
# Source: kiali-server/templates/role-viewer.yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: kiali-viewer
  labels:
    helm.sh/chart: kiali-server-1.57.0
    app: kiali
    app.kubernetes.io/name: kiali
    app.kubernetes.io/instance: kiali
    version: "v1.57.0"
    app.kubernetes.io/version: "v1.57.0"
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/part-of: "kiali"
rules:
- apiGroups: [""]
  resources:
  - configmaps
  - endpoints
  - pods/log
  verbs:
  - get
  - list
  - watch
- apiGroups: [""]
  resources:
  - namespaces
  - pods
  - replicationcontrollers
  - services
  verbs:
  - get
  - list
  - watch
- apiGroups: [""]
  resources:
  - pods/portforward
  verbs:
  - create
  - post
- apiGroups: ["extensions", "apps"]
  resources:
  - daemonsets
  - deployments
  - replicasets
  - statefulsets
  verbs:
  - get
  - list
  - watch
- apiGroups: ["batch"]
  resources:
  - cronjobs
  - jobs
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - networking.istio.io
  - security.istio.io
  - extensions.istio.io
  - telemetry.istio.io
  - gateway.networking.k8s.io
  resources: ["*"]
  verbs:
  - get
  - list
  - watch
- apiGroups: ["apps.openshift.io"]
  resources:
  - deploymentconfigs
  verbs:
  - get
  - list
  - watch
- apiGroups: ["project.openshift.io"]
  resources:
  - projects
  verbs:
  - get
- apiGroups: ["route.openshift.io"]
  resources:
  - routes
  verbs:
  - get
- apiGroups: ["authentication.k8s.io"]
  resources:
  - tokenreviews
  verbs:
  - create
...
```

If you want the read-write permissions, you can do:

```
helm template --version 1.57.0 --show-only templates/role.yaml kiali-server kiali/kiali-server
```

So we wouldn't have to change the docs as we change the Roles, we can leave just that helm template command and the user can generate the yaml that they want, for whatever version they want.

israel-hdez:
> If you want the read-write permissions, you can do:
> 
> ```
> helm template --version 1.57.0 --show-only templates/role.yaml kiali-server kiali/kiali-server
> ```


The problem with that, and the motivation for being explicit in the docs, is that you don't need all those privileges. As mentioned in the new writing, you just need this minimal privs:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: kiali-namespace-authorization
rules:
- apiGroups: [""]
  resources:
  - namespaces
  - pods/log
  verbs:
  - get
- apiGroups: ["project.openshift.io"] # Only if you are using OpenShift
  resources:
  - projects
  verbs:
  - get
```

With this minimal role, Kiali will let you view everything else.

Looking forward, since the kube cache is now mandatory in Kiali, I think it is expected that this minimal role will remain stable, even if the KialiSA needs to get more privileges.

This is true for **read** privileges.

For **write** privileges, you _do_ need to give explicit privs, but the helm charts do not currently have a way to generate a Role or ClusterRole only for write privs.

israel-hdez:
What I want to communicate in the new docs, is that **you can't** fine-tune read privileges in Kiali. If you can do `kubectl get namespace foo`, Kiali will let you read everything in that namespace even if you don't have any other privilege. So, it doesn't make sense to take the KialiSA role as a template.

jmazzitelli:
@israel-hdez OK, I think I get it now. You are documenting the USER'S Role which doesn't need all the read-permissions the Kiali SA has (essentially, you are saying the USER will get the read-permissions that the Kiali SA has, regardless of what the USER'S read permissions are). Correct?

israel-hdez:
> (essentially, you are saying the USER will get the read-permissions that the Kiali SA has, regardless of what the USER'S read permissions are). Correct?

Yes! "So-so". Always limited to what the Kiali API lets you get -- e.g The Kiali SA may have privs to read some secrets, but Kiali does not have an API that would let you read them :-)

content/en/docs/Configuration/authentication/openid.md
```diff
@@ -14,9 +14,8 @@ third-party system.
 
 If your
 [Kubernetes cluster is also integrated with your OpenId provider](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens),
-then Kiali's `openid` strategy can offer role-based access control (RBAC) through the
-[Kubernetes authorization mechanisms](https://kubernetes.io/docs/reference/access-authn-authz/rbac/). See the
-[RBAC documentation]({{< relref "../rbac" >}}) for more details.
+then Kiali's `openid` strategy can offer 
+[namespace access control]({{< relref "../rbac" >}}) for more details.
```
jmazzitelli:
```suggestion
[namespace access control]({{< relref "../rbac" >}}).
```

content/en/docs/Configuration/rbac.md
```diff
@@ -1,120 +1,214 @@
...
-description: "Configuring RBAC based on the authentication strategy."
+title: "Namespace access control"
+description: "Configuring per-user authorized namespaces."
 ---
 
 ## Introduction
 
-Kiali supports role-based access control (RBAC) when using any authentication strategy other than `anonymous`.
+In authentication strategies other than `anonymous` Kiali supports limiting the
+namespaces that are accessible in a per-user basis. The `anonymous`
```
jmazzitelli:
```suggestion
namespaces that are accessible on a per-user basis. The `anonymous`
```

content/en/docs/Configuration/rbac.md
```diff
@@ -1,120 +1,214 @@
...
+{{% alert color="warning" %}}
+Kiali is going to reject login to users that aren't authorized to see any namespace.
+{{% /alert %}}
 
-Please read your cluster RBAC documentation to learn how to assign privileges.
+## Granting access to namespaces
 
-## Minimum required privileges to login
+In general, Kiali will give _read_ access to namespaces where the logged in
+user is granted to _"GET"_ its definition -- i.e. the user is allowed to do a
```
jmazzitelli:
```suggestion
user is allowed to _"GET"_ its definition -- i.e. the user is allowed to do a
```

content/en/docs/Configuration/rbac.md
```diff
@@ -1,120 +1,214 @@
...
-$ kubectl create rolebinding john-binding --clusterrole=kiali --user="john" --namespace=myApp
-$ oc adm policy add-role-to-user kiali john -n myApp # For OpenShift clusters
-```
+## Granting write privileges to namespaces
 
-But if you need to assign cluster-wide privileges, you could run either:
+Changing resources in the cluster can be a sensitive operation. Because of
+this, the logged in user will need to be given the needed privileges to perform
+any updates through Kiali. The following `ClusterRole` contains all write
+privileges that may used in Kiali:
```
jmazzitelli:
```suggestion
privileges that may be used in Kiali:
```

content/en/docs/Configuration/rbac.md
```diff
@@ -1,120 +1,214 @@
...
-based off the privileges in Kiali's `ClusterRole` and remove the privileges you
-want to ban. You must understand that some Kiali features may not work properly
-because of the reduced privilege set.
 
-{{% alert color="warning" %}}
-Do not edit the `kiali` or the `kiali-viewer`
-`ClusterRoles` because they are bound to the Kiali ServiceAccount and editing
-them may lead to Kiali not working properly.
+{{% alert color="info" %}}
+If needed, you can reduce the set of write privileges to prevent users changing
```
jmazzitelli:
```suggestion
If needed, you can reduce the set of write privileges to prevent users from changing
```

jmazzitelli (changes requested):
just some minor things

israel-hdez:
@jmazzitelli I applied all your suggestions.

jmazzitelli (approved):

PR #633, opened 2023-02-08
Generate Istio compatibility table from YAML

sspaeth-r7:
In reference to earlier today's issue [kiali/kiali#5823](https://github.com/kiali/kiali/issues/5823), while I was closing out related tabs I noticed the site is written in Hugo. That made me think of doing something like this.

This PR brings back a YAML file for the Istio compatibility matrix, but makes it the source of truth and automatically generates the table in the documentation. I see this as the best of both worlds, because contributors only have to update the matrix in one place, while it's still possible (and feasible) to build automation around.

It also makes a couple of changes to the README in response to problems I encountered during local setup. Running the command as is caused this error:
```
make: *** No rule to make target `make'.  Stop.
```
But the suggested command works fine.

If this is something y'all would be okay with, I'm happy to make the same changes on the other two compatibility matrices. I was originally wanting to make a reusable, single shortcode for a YAML-generated table, but due to the differences in header names, their sizes, and property names, I figured I'd end up either polluting the YAML files, or would have too many arguments when calling the shortcode. I've never used Hugo though, so if the way I implemented it is poor, I'm happy to correct and define the header in each YAML file and then read that dynamically in the shortcode.


Xunzhuo (approved):

Xunzhuo:
https://deploy-preview-633--kiali.netlify.app/docs/installation/installation-guide/prerequisites/

jshaughn (approved):
@sspaeth-r7 , thanks for the contribution and the lesson in how to create shortcodes :-)  This works fine for me and I assume it's better for you.

sspaeth-r7:
No problem, appreciate the openness to this idea!

PR #644, opened 2023-03-29
fix confusing doc regarding operator helm settings with a better and correct example

jmazzitelli:
https://deploy-preview-644--kiali.netlify.app/docs/installation/installation-guide/install-with-helm/#install-with-operator


jshaughn (approved):
LOK

PR #656, opened 2023-05-23
talk about the openshift auth strategy config settings

jmazzitelli:
there are some openshift config options now - change the docs to reflect this
netlify: https://deploy-preview-656--kiali.netlify.app/docs/configuration/authentication/openshift/


jshaughn (approved):
Seems like a good compromise between documenting and removing.

PR #657, opened 2023-05-23
Update docs with multi-cluster info

nrfox:
Relates to kiali/kiali#5928


jshaughn:
@nrfox This PR is in Draft, should we wait to review?

nrfox:
> @nrfox This PR is in Draft, should we wait to review?

I opened it as a draft because the current OIDC implementation for multi-cluster may not make it into the alpha but please review the content that is there.

Makefile
```diff
@@ -71,4 +71,4 @@ URL_IGNORE:=$(URL_IGNORE)$(NEW_URLS)
 ## validate-site: Builds the site and validates the pages. This is used for CI
 .PHONY: validate-site
 validate-site: build-hugo
-	${DORP} run -t -i --rm -v "$(shell pwd)":/site:z -w /site ${KIALI_HUGO_IMAGE} /bin/bash -c "npm prune && npm config set fetch-retry-mintimeout 20000 && npm config set fetch-retry-maxtimeout 120000 && hugo && htmlproofer --typhoeus '{\"connecttimeout\": 30, \"timeout\": 30}' --hydra='{\"max_concurrency\": 6}' --allow-hash-href --allow-missing-href --ignore-empty-alt --ignore-missing-alt --no-check-external-hash --no-check-internal-hash --no-enforce-https --ignore-urls \"${URL_IGNORE}\" ./public"
\ No newline at end of file
+	${DORP} run -t -i --rm -v "$(shell pwd)":/site:z -w /site ${KIALI_HUGO_IMAGE} /bin/bash -c "npm prune && npm config set fetch-retry-mintimeout 20000 && npm config set fetch-retry-maxtimeout 120000 && hugo && htmlproofer --typhoeus '{\"connecttimeout\": 30, \"timeout\": 30}' --hydra='{\"max_concurrency\": 6}' --allow-hash-href --allow-missing-href --ignore-empty-alt --ignore-missing-alt --no-check-external-hash --no-check-internal-hash --no-enforce-https --ignore_status_codes "302" --ignore-urls  \"${URL_IGNORE}\" ./public"
```
jshaughn: I guess this fixes that redirect problem?  If so, thank you!

content/en/docs/Configuration/multi-cluster.md
```diff
@@ -0,0 +1,67 @@
...
+
+Kiali has [experimental support for Istio multi-cluster installations]({{< relref "../Features/multi-cluster" >}}).
+
+Before proceeding with the setup, ensure your deployment meets the following requirements.
+
+## Requirements
+
+1. Primary-remote deployment
+
+Only the primary-remote deployment is currently supported.
```
jshaughn: I think we should get away from the "primary-remote", or "multi-primary" terminology.  You'll notice that Istio has no mention of these terms anymore, in their Deployment Models section.

I think maybe we should just emphasize our own Deployment models.  Which at the moment is (I think):
- One Kiali per control plane (istiod)
  - No support for external control planes
  - Kiali and istiod on the same cluster
- One Kiali per mesh
  - Mesh can span multiple clusters
- One or more data planes
- One metric store (Prometheus or Thanos)
  - multiple clusters require unified metrics
- One trace store
  - multiple clusters require unified traces

content/en/docs/Features/multi-cluster.md
```diff
@@ -3,22 +3,46 @@ title: "Multi-cluster Deployment"
...
-[multi-network](https://istio.io/latest/docs/ops/deployment/deployment-models/#multiple-networks) approach.  The only strict rule is that within a mesh service names are unique. A non-basic mesh deployment generally involves multiple clusters.  See [installation instructions](https://istio.io/docs/setup/install/multicluster/) for more detail on installing advanced mesh deployments.
+[multi-network](https://istio.io/latest/docs/ops/deployment/deployment-models/#multiple-networks) approach. The only strict rule is that within a mesh service names are unique. A non-basic mesh deployment generally involves multiple clusters. See [installation instructions](https://istio.io/docs/setup/install/multicluster/) for more detail on installing advanced mesh deployments.
+
+For multi-cluster deployments, Kiali provides a single place to manage your mesh across clusters. Support is still experimental and only primary-remote deployments are currently supported but support for more deployment models is in progress.
```
jshaughn: See other comment, I think we should get away from "primary-remote" terminology.

content/en/docs/Features/multi-cluster.md
```diff
@@ -3,22 +3,46 @@ title: "Multi-cluster Deployment"
...
+## Previous mesh support
 
-Each new box type supports selection and will provide a side-panel summary of traffic.  Below we see a Bookinfo traffic graph for when Bookinfo services are deployed across the Kukulcan and Tzotz clusters. The Kukulcan cluster box is selected. Because Kukulcan is the Kiali home cluster, you see traffic from the perspective of Kukulcan. Note that you do not see the internal traffic on Tzotz (the requests from Reviews to Ratings service).  To see traffic from the Tzotz cluster point of view, you would open a Kiali session on Tzotz, assuming you have privileges.
+Kiali has had experimental support for multi-cluster meshes since v1.29. Previously this support required multiple Kialis to run in each Kubernetes cluster and only offered a partial view of the entire mesh. Now Kiali lets you view your entire mesh from a single Kiali deployment. Much of the previous support for multi-cluster has been retained but links to external Kialis and autodiscovery of external Kialis has been removed.
```
jshaughn: I feel like we should have maintained support for that model of "linked" Kiali's for when the user has multiple control planes with a Kiali on each, and may want to jump from one to the other.

jshaughn (commented):
A couple of comments.  Additionally, I wonder if this MC documentation should also provide a section talking about how `query_scope` can be used (for metrics and traces), when dealing with unified stores.

Makefile
nrfox: Yeah it just ignores redirects. This might not be what we want since redirects could lead to a page that we weren't originally trying to link to but if there's a redirect then at least the link won't be broken.

content/en/docs/Configuration/multi-cluster.md
nrfox: If you follow the install directions for multi-cluster, each configuration still uses the "primary" and "remote" terminology:
Screenshot from 2023-05-25 08-06-35

I think it's nice because it succinctly describes a specific configuration and for folks following the multi-cluster istio install docs it would provide some continuity but simply saying "single controlplane" or "multi controlplane" is probably more clear especially if you don't know what "primary" and "remote" are. Perhaps we can use both?

> I think maybe we should just emphasize our own Deployment models

Agreed I think it'd make it more clear to lay it out this way.

content/en/docs/Features/multi-cluster.md
nrfox: > I feel like we should have maintained support for that model of "linked" Kiali's for when the user has multiple control planes with a Kiali on each, and may want to jump from one to the other.

We still can. Technically we haven't removed support for this yet even though it has never graduated from "experimental". It's also possible this feature has degraded some since the other multi-cluster work has been underway and it has been largely untested. The goal is to support a single Kiali for multiple control planes but users may still want the option of a kiali per controlplane. I'm not sure if this is something users really want though since Kiali has had this support for awhile but what many users have asked for is a single unified view. If we are going to keep the Kiali per controlplane support then we need to add some automated testing around it as we will for the other deployment models.

content/en/docs/Configuration/multi-cluster.md
jshaughn: Ah, OK, I didn't see how they laid out the Install section.  So I retract what I said.  When I looked at the Deployment Models page I didn't see it.  

Is it true that we don't support Multi-Primary?  We've always said that we do, using a Kiali per control plane.  So I'm a bit afraid that it looks like we're giving up support for Multi-Primary.  Minimally it would seem like we could support it like we always have, just that each Kiali may have a limited view (like in Federation).  But could it not also be like two Primary-Remote installs?

nrfox: > Is it true that we don't support Multi-Primary?

You can always run an independent Kiali per controlplane. The support we have now for multi-primary is to provide a link to each independent Kiali. But that's not really the same as having a unified Kiali. After discussing this a little further I think it'd be good to clarify the "unified view" Kiali from just running independent Kialis.

content/en/docs/Configuration/multi-cluster.md
```diff
@@ -0,0 +1,93 @@
...
+
+1. **Create a remote kubeconfig secret.** You can use [this script](https://github.com/kiali/kiali/blob/master/hack/istio/multicluster/kiali-prepare-remote-cluster.sh) to simplify this process for you. Running this script will:
+
+   - Create a Kiali Service Account in the remote cluster.
+   - Create a role/role-binding for this service account in the remote cluster.
+   - Create a kubeconfig file and save this as a secret in the namespace where Kiali is deployed. The Kiali operator will auto-detect the secret and mount it into the Kiali pod.
+
+   In order to run this script you will need adaquete permissions configured in your local kubeconfig for both the cluster Kiali is deployed in and the remote cluster. You will need to repeat this step for each remote cluster.
```
josunect: I think it should be "adequate" ?

content/en/docs/Features/multi-cluster.md
```diff
@@ -1,24 +1,56 @@
...
-## Graph View: Cluster and Namespace Boxing
+![Multi-cluster mesh discovery](/images/documentation/features/multi-cluster-mesh-v1.29.png "Multi-cluster mesh discovery")
+
+## Unified Multi-cluster configuration
+
+Unlike single-cluster configurations, unified configurations show list/details pages across all clusters.
+
+### List Views: Aggregated mesh view
+
+With a unified Kiali configuration, you can view all apps, workloads, services, and Istio config in your mesh from a single place.
```
josunect: Should we add a note indicating that istio configs are available in both clusters but just writes are supported for the local cluster?

content/en/docs/Features/multi-cluster.md
```diff
@@ -1,24 +1,56 @@
...
 
 ## List View: Mesh Discovery
-Kiali will attempt to discover the clusters configured in the mesh.  And it will identify the *home* cluster, meaning the cluster on which it is installed and from which it presents its traffic, traces and Istio config.  In the following example there are two clusters defined in the mesh, _Kukulcan_ and _Tzotz_.  Kukulcan is identified as the home cluster in three places: the browser tab (not shown), the masthead, and with a star icon in the clusters list:
 
-![Mesh list view](/images/documentation/features/multi-cluster-mesh-v1.29.png "Mesh list view")
+Kiali will show cluster information for all clusters in the mesh. In single-cluster configurations, Kiali will attempt to discover the clusters configured in the mesh. It will identify the home cluster, meaning the cluster on which it is installed and from which it presents its traffic, traces and Istio config. In the following example there are two clusters defined in the mesh, Kukulcan and Tzotz. Kukulcan is identified as the home cluster in three places: the browser tab (not shown), the masthead, and with a star icon in the clusters list:
```
josunect: Maybe we can update the image here? 
So the example cluster names are the same. Also here, it is included the Kiali column.

image

content/en/docs/Features/multi-cluster.md
nrfox: Yeah that'd be good to add

content/en/docs/Features/multi-cluster.md
nrfox: Talking about this further, generating links to external kialis should work _without_ providing access to remote clusters. By adding access to remote clusters, you are effectively enabling "multi-cluster mode" for Kiali. We can keep the existing behavior of generating links to external kialis with the secrets but by adding the secrets you will also get a "unified" view. In the future, Kiali should support generating the links without needing to grant access to remote clusters by simply providing some config values.

content/en/docs/Configuration/multi-cluster.md
```diff
@@ -0,0 +1,63 @@
...
+
+![Kiali multi-cluster](/images/documentation/configuration/multi-cluster.png)
+
+Before proceeding with the setup, ensure you meet the requirements.
+
+#### Requirements
+
+1. **Primary-remote istio deployment.** Only the primary-remote istio deployment is currently supported.
+
+2. **Aggregated metrics and traces.** Kiali needs a single endpoint where it can consume aggregated metrics/traces across all clusters. There are many ways to aggregate metrics/traces such as prometheus federation or using OTEL collector pipelines but setting these up are outside of the scope of Kiali.
```
jshaughn: This sounds like you need a single endpoint that serves up both metrics and traces.  Maybe it should be more clear that it's one endpoint for metrics and one endpoint for traces?

content/en/docs/Configuration/multi-cluster.md
```diff
@@ -0,0 +1,63 @@
...
+
+3. **Anonymous or OpenID authentciation strategy.** The unified multi-cluster configuration currently only supports anonymous or OpenID [authentication strategies]({{< relref "../Configuration/authentication" >}}). In addition, currently support for OpenID across clusters varies by provider.
```
jshaughn:
```suggestion
3. **Anonymous or OpenID authentication strategy.** The unified multi-cluster configuration currently only supports anonymous or OpenID [authentication strategies]({{< relref "../Configuration/authentication" >}}). In addition, current support varies by provider for OpenID across clusters.
```

content/en/docs/Configuration/multi-cluster.md
```diff
@@ -0,0 +1,63 @@
...
+
+#### Setup
+
+The unified Kiali multi-cluster setup requires the Kiali Service Account (SA) to have read access to each Kubernetes cluster in the mesh. This is separate from the user credentials that are required when a user logins to Kiali. The user credentials are used to check user access to a namespace and to perform write operations. In anonymous mode, the Kiali SA is used for all operations and write access is also required. To give the Kiali SA access to each remote cluster, a kubeconfig with credentials needs to be created and mounted into the Kiali pod. While the location of Kiali in relation to the controlplane and dataplane may change depending on your istio deployment model, the requirements will remain the same.
```
jshaughn:
```suggestion
The unified Kiali multi-cluster setup requires the Kiali Service Account (SA) to have read access to each Kubernetes cluster in the mesh. This is separate from the user credentials that are required when a user logs into Kiali. The user credentials are used to check user access to a namespace and to perform write operations. In anonymous mode, the Kiali SA is used for all operations and write access is also required. To give the Kiali SA access to each remote cluster, a kubeconfig with credentials needs to be created and mounted into the Kiali pod. While the location of Kiali in relation to the controlplane and dataplane may change depending on your istio deployment model, the requirements will remain the same.
```

content/en/docs/Configuration/multi-cluster.md
```diff
@@ -0,0 +1,63 @@
...
+
+1. **Create a remote kubeconfig secret.** You can use [this script](https://github.com/kiali/kiali/blob/master/hack/istio/multicluster/kiali-prepare-remote-cluster.sh) to simplify this process for you. Running this script will:
+
+   - Create a Kiali Service Account in the remote cluster.
+   - Create a role/role-binding for this service account in the remote cluster.
+   - Create a kubeconfig file and save this as a secret in the namespace where Kiali is deployed. The Kiali operator will auto-detect the secret and mount it into the Kiali pod.
+
+   In order to run this script you will need adequate permissions configured in your local kubeconfig for both the cluster Kiali is deployed in and the remote cluster. You will need to repeat this step for each remote cluster.
```
jshaughn:
```suggestion
   In order to run this script you will need adequate permissions configured in your local kubeconfig, for both the cluster on which Kiali is deployed and the remote cluster. You will need to repeat this step for each remote cluster.
```

content/en/docs/Configuration/multi-cluster.md
```diff
@@ -0,0 +1,63 @@
...
+
+   ```
+   curl -L -o kiali-prepare-remote-cluster.sh https://raw.githubusercontent.com/kiali/kiali/master/hack/istio/multicluster/kiali-prepare-remote-cluster.sh
+   chmod +x kiali-prepare-remote-cluster.sh
+   ./kiali-prepare-remote-cluster.sh --kiali-cluster-context east --remote-cluster-context west --view-only false
+   ```
+
+Adding remote kubeconfig secrets to Kiali effectively puts Kiali in "multi-cluster" mode and Kiali will begin using those credentials to communicate with the other clusters in the mesh.
+
+2. Optional - **Configure tracing with cluster ID.** By default, traces do not include their cluster name in the trace attributes however this can be added using the istio telemetry API.
```
jshaughn: attributes -> tags?

content/en/docs/Configuration/multi-cluster.md
```diff
@@ -0,0 +1,63 @@
...
+  -  customTags:
+      cluster:
+        environment:
+          name: "ISTIO_META_CLUSTER_ID"
+EOF
+```
+
+3. Optional - **Configure user access in your OIDC provider.** When using anonymous mode, the Kiali SA credentials will be used to display mesh info to the user. When not using anonymous mode, Kiali will check the user's access to each configured cluster's namespace before showing the user any resources from that namespace. Please refer to your OIDC provider's instructions for configuring user access to a kube cluster for this.
+
+4. Optional - **Narrow metrics to mesh clusters.** If your unified metrics store also contains data for clusters outside of your mesh, you can limit which clusters Kiali will query for by setting the [query_scope](/docs/configuration/kialis.kiali.io#.spec.external_services.custom_dashboards.prometheus.query_scope) configuration.
```
jshaughn:
```suggestion
4. Optional - **Narrow metrics to mesh.** If your unified metrics store also contains data outside of your mesh, you can limit which metrics Kiali will query for by setting the [query_scope](/docs/configuration/kialis.kiali.io#.spec.external_services.custom_dashboards.prometheus.query_scope) configuration.
```

content/en/docs/Features/multi-cluster.md
```diff
@@ -1,24 +1,56 @@
 ---
-title: "Multi-cluster Deployment"
+title: "Multi-cluster"
 description: "Advanced Mesh Deployment and Multi-cluster support."
 ---
 
-A basic Istio mesh deployment has a single control plane with a single data plane, deployed on a single Kubernetes cluster. But Istio supports a variety of advanced 
+{{% alert color="warning" %}}
+Support for multi-cluster deployments is currently **experimental** and subject to change. It is not suitable for production environments.
```
jshaughn: Are we sure about saying it's not suitable for production?  Kiali is not usually critical path, I wouldn't want someone to skip using Kiali because they think it's going to break.  Not a strong opinion from me, just wondering if "experimental" is sufficient.

content/en/docs/Features/multi-cluster.md
```diff
@@ -1,24 +1,56 @@
...
+[multi-network](https://istio.io/latest/docs/ops/deployment/deployment-models/#multiple-networks) approach. The only strict rule is that within a mesh service names are unique. A non-basic mesh deployment generally involves multiple clusters. See [installation instructions](https://istio.io/docs/setup/install/multicluster/) for more detail on installing advanced mesh deployments.
+
+A single Kiali install can currently work with at most one mesh, one istiod, one metric store and one trace store but it can be configured for "single cluster" or "multi cluster". All clusters must be part of the same mesh, have data planes controlled by the single istiod (control plane), and report to the same metric and trace store, whether directly or via some sort of aggregator. See the [multi-cluster configuration page]({{< relref "../Configuration/multi-cluster" >}}) for more information on requirements.
```
jshaughn:
```suggestion
A single Kiali install can currently work with at most one mesh, one istiod, one metric store and one trace store but it can be configured for "single cluster" or "multi-cluster". All clusters must be part of the same mesh, have data planes controlled by the single istiod (control plane), and report to the same metric and trace store, whether directly or via some sort of aggregator. See the [multi-cluster configuration page]({{< relref "../Configuration/multi-cluster" >}}) for more information on requirements.
```

content/en/docs/Features/multi-cluster.md
jshaughn: I think we captured this "ask" as a new configuration item, so resolving this thread.

jshaughn (changes requested):
Just some grammar stuff, a couple of minor questions.

content/en/docs/Configuration/multi-cluster.md
nrfox: hmmmm I think these are called "tags" by jaeger and "attributes" by [otel](https://opentelemetry.io/docs/concepts/signals/traces/#attributes) and they seem to be called "tags" by [istio](https://istio.io/latest/docs/reference/config/telemetry/#Tracing-CustomTag) so I'll go with that.

content/en/docs/Features/multi-cluster.md
nrfox: I'm good with keeping the "experimental" label but removing the production warning. The only "dangerous" thing I see is in giving Kiali remote access to clusters. However this can be read only and should be gated by RBAC if Kiali is not running in anonymous mode. Anonymous mode does not give you any protections in single cluster either so nothing has changed there.

jshaughn (approved):
All good from my side!

jshaughn:
One additional comment, if everything applies to v1.69 then we should also backport this to `current`.

jmazzitelli (approved):
lgtm

PR #663, opened 2023-06-21
Update arch diagram

### LINKED ISSUE #5376: Architecture is outdated - direct istio dependency
The [Kiali architecture section and diagram](https://kiali.io/docs/architecture/architecture/#kiali-architecture) both suggest that istio/istiod are indirect dependencies but now Kiali does directly rely on and communicate with istio/istiod to gather information about workloads and services.

nrfox:
Couldn't find the old image anywhere to update so I added a new image that shows that istiod is a required dependency but also adds a note in the description that it can be disabled. Also changes the Cluster API --> Kube API because IMO Cluster API is confusing.

Fixes kiali/kiali#5376


content/en/docs/Architecture/architecture.md
```diff
@@ -49,8 +49,8 @@ It provides and controls the service mesh. Kiali and Istio are installed
 separately.
 
 Kiali needs to retrieve Istio data and configurations, which are exposed
-through Prometheus and the cluster API. This is the reason the diagram shows a
-dashed line: to denote an indirect dependency.
+through Prometheus, the Kubernetes API, and istiod. For environments where
```
leandroberetta, line 52: ultra minor detail but the comma before "and istiod" should be ommited right?

leandroberetta (commented):
Just a ultra minor comment, but looks good.

content/en/docs/Architecture/architecture.md
nrfox: The [oxford comma](https://en.wikipedia.org/wiki/Serial_comma) is a contentious subject actually. I've always put one on the end of a list but I'm certainly no authority on the subject. I'm happy to remove it and break with my tradition.

nrfox: It looks like other places on the same page in the docs use it as well.
> Kiali queries the Kubernetes API to retrieve, for example, definitions for
> namespaces, services, deployments, pods, and other entities. 
> ...
> 
> The Kubernetes API is also queried to retrieve Istio configurations like virtual
> services, destination rules, route rules, gateways, and quotas.

How do you feel about keeping it for consistency's sake?

leandroberetta: Feels ok!

leandroberetta (approved):
LG!

PR #681, opened 2023-08-02
Multicluster tutorial

leandroberetta:
Multi cluster tutorial. There are things we could add, like integration with Keycloak to show permissions. Also, this is just for primary and remote, so I see in the near future adding the multi primary case or doing another one for that.

netlify link: https://deploy-preview-681--kiali.netlify.app/docs/tutorials/multicluster/

Resolves https://github.com/kiali/kiali/issues/6306

PR #689, opened 2023-08-22
Mc tutorial suggestions

nrfox:
Follow up from [review suggestions](https://github.com/kiali/kiali.io/pull/681#pullrequestreview-1571674509). Mostly just adds a bit more info about the multicluster environment and tweaks some sentences here and there. My editor also applied some autolinting.

Also fixes some links that were causing build failures to get CI passing again.


leandroberetta (approved):
Nick, these changes looks fine. Thanks for the review.

PR #693, opened 2023-09-05
State minimum supported Helm version

### LINKED ISSUE #6566: document minimum helm version
We have had several community members reports issues with the helm charts. It is because they are not using Helm 3.10.

Our docs are not firm enough - we need to tell people Helm 3.10 is the minimum version and anything older will not work.

See: https://kiali.io/docs/installation/installation-guide/install-with-helm/#introduction

Where it says, `The Kiali Helm Charts have been tested only against Helm version 3.10. There is no guarantee that previous versions will work.`

That needs to say something more definitive like `The Kiali Helm Charts have been tested only against Helm version 3.10.  Any previous versions will not work and are not supported.`

jshaughn:
Fixes https://github.com/kiali/kiali/issues/6566

https://deploy-preview-693--kiali.netlify.app/docs/installation/installation-guide/install-with-helm/#introduction

### DIFF
--- a/content/en/docs/Installation/installation-guide/install-with-helm.md
+++ b/content/en/docs/Installation/installation-guide/install-with-helm.md
@@ -32,7 +32,7 @@ Make sure you have the `helm` command available by following the
 [Helm installation docs](https://helm.sh/docs/intro/install/).
 
 {{% alert color="warning" %}}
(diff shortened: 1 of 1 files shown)


jmazzitelli (approved):
bueno

PR #715, opened 2023-11-06
Include alternative to set up cluster tracing tag in multi cluster 

josunect:
Include alternative to set up cluster tracing tags in multi cluster without using Telemetry API.

### DIFF
--- a/content/en/docs/Configuration/multi-cluster.md
+++ b/content/en/docs/Configuration/multi-cluster.md
@@ -60,14 +60,12 @@ As an alternative, it can be added in the Istio tracing config in the primary cl
 
 ```
 meshConfig:
-  accessLogFile: /dev/stdout
-    defaultConfig:
-      tracing:
-        sampling: 100
-        custom_tags:
-          cluster:
-            environment:
-              name: ISTIO_META_CLUSTER_ID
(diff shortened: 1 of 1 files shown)


content/en/docs/Configuration/multi-cluster.md
```diff
@@ -56,6 +56,20 @@ spec:
...
+```
+meshConfig:
+  accessLogFile: /dev/stdout
+    defaultConfig:
+      tracing:
+        sampling: 100
+        custom_tags:
+          cluster:
+            environment:
+              name: ISTIO_META_CLUSTER_ID
```
nrfox: I think the indentation is off here and we should only include what is needed to add the cluster id:

```suggestion
meshConfig:
  defaultConfig:
    tracing:
      custom_tags:
        cluster:
          environment:
            name: ISTIO_META_CLUSTER_ID
```

nrfox (changes requested):

content/en/docs/Configuration/multi-cluster.md
josunect: Thanks, @nrfox ! Good point.

nrfox (approved):

PR #720, opened 2023-11-20
remove extra space in advanced install options

cuichenli:


aljesusg (approved):
LGFM

jmazzitelli:
HA! Good catch.. how did we miss that :)

PR #730, opened 2023-12-06
medium rss changed - need to fix the docs that pull it in

jmazzitelli:
netlify: https://deploy-preview-730--kiali.netlify.app/#td-block-3


jshaughn (approved):
BYC

PR #759, opened 2024-02-19
Remove base64 hint in kiali-cabundle ConfigMap example

steadyk:
Currently there is a hint to provide base64 encoded certificates in the kiali-cabundle ConflgMap when using an own root CA certificate.

However this doesn't work, since Kiali tries to read the certificate in plaintext (see [openID-auth-controller](https://github.com/kiali/kiali/blob/82719968b06f7d69e5955efe08b6a3d59f1e8b8f/business/authentication/openid_auth_controller.go#L923)).

This PR just removes the base64 hint.


jshaughn (approved):
Thanks, @steadyk.

PR #765, opened 2024-03-19
Improve Install via Helm.

chickenchickenlove:
### Background
Hi, I'm new in `kiali` and this is my first PR! 
When I first tried to install the `Kiali` operator and access the `Kiali UI`, I encountered a problem where the token could not be found, and it was difficult to find a solution through searching the documentation. 
The problem occurred with the commands below.
```shell
$ helm install \
    --set cr.create=true \
    --set cr.namespace=istio-system \
    --namespace kiali-operator \
    --create-namespace \
    kiali-operator \
    kiali/kiali-operator
```
Thus, IMO, it will be better to add `--set cr.auth.strategy="anonymous"` for newbie in `kiali` to started with `kiali` easily.

### Changes
- Add `--set cr.auth.strategy="anonymous"` to install via helm page.
- Add some comments to describe accessing kiali UI at install via helm page.

### DIFF
--- a/content/en/docs/Installation/installation-guide/install-with-helm.md
+++ b/content/en/docs/Installation/installation-guide/install-with-helm.md
@@ -84,7 +84,7 @@ cr.namespace=istio-system` flags instructs to create a Kiali CR in the
 the Kiali operator starts, it will process it to deploy Kiali. After Kiali has started,
 you can access Kiali UI through 'http://localhost:20001' by executing 
 `kubectl port-forward service/kiali -n istio-system 20001:20001`
-because of `--set cr.auth.strategy="anonymous"`.
+because of `--set cr.auth.strategy="anonymous"`. But realize that anonymous mode will allow anyone to be able to see and use Kiali. If you wish to require users to authenticate themselves by logging into Kiali, use one of the other [auth strategies](/configuration/authentication/).
 
 The Kiali Operator Helm Chart is configurable. Check available options and default values by running:
 
--- a/content/en/docs/Installation/installation-guide/install-with-helm.md
+++ b/content/en/docs/Installation/installation-guide/install-with-helm.md
@@ -84,7 +84,7 @@ cr.namespace=istio-system` flags instructs to create a Kiali CR in the
 the Kiali operator starts, it will process it to deploy Kiali. After Kiali has started,
 you can access Kiali UI through 'http://localhost:20001' by executing 
 `kubectl port-forward service/kiali -n istio-system 20001:20001`
-because of `--set cr.auth.strategy="anonymous"`. But realize that anonymous mode will allow anyone to be able to see and use Kiali. If you wish to require users to authenticate themselves by logging into Kiali, use one of the other [auth strategies](/configuration/authentication/).
+because of `--set cr.auth.strategy="anonymous"`. But realize that anonymous mode will allow anyone to be able to see and use Kiali. If you wish to require users to authenticate themselves by logging into Kiali, use one of the other [auth strategies](/docs/configuration/authentication/).
 
 The Kiali Operator Helm Chart is configurable. Check available options and default values by running:
 
--- a/content/en/docs/Installation/installation-guide/install-with-helm.md
+++ b/content/en/docs/Installation/installation-guide/install-with-helm.md
@@ -69,6 +69,7 @@ command:
 $ helm install \
     --set cr.create=true \
     --set cr.namespace=istio-system \
+    --set cr.auth.strategy="anonymous" \
     --namespace kiali-operator \
     --create-namespace \
     kiali-operator \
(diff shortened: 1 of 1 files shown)


content/en/docs/Installation/installation-guide/install-with-helm.md
```diff
@@ -80,7 +81,10 @@ create the `kiali-operator` namespace (if needed), and deploy the Kiali
 operator on it.  The `--set cr.create=true` and `--set
 cr.namespace=istio-system` flags instructs to create a Kiali CR in the
 `istio-system` namespace. Since the Kiali CR is created in advance, as soon as
-the Kiali operator starts, it will process it to deploy Kiali.
+the Kiali operator starts, it will process it to deploy Kiali. After Kiali has started,
+you can access Kiali UI through 'http://localhost:20001' by executing 
+`kubectl port-forward service/kiali -n istio-system 20001:20001`
+because of `--set cr.auth.strategy="anonymous"`.
```
jmazzitelli:
```suggestion
because of `--set cr.auth.strategy="anonymous"`. But realize that anonymous mode will allow anyone to be able to see and use Kiali. If you wish to require users to authenticate themselves by logging into Kiali, use one of the other [auth strategies](/configuration/authentication/).
```

jmazzitelli (changes requested):
This is OK. The reason we did not specify anonymous setting before is because this will open up the Kiali UI to anyone. So it isn't secure. Because of this, just add a blurb at the end to call this out to the reader so they are aware running that command is not secure, and if they want to make it secure, there are options to do so. See my suggestion (make sure that link works :).

jmazzitelli:
The netlify link to go to in order to see the change is found here: https://deploy-preview-765--kiali.netlify.app/docs/installation/installation-guide/install-with-helm/#install-with-operator

content/en/docs/Installation/installation-guide/install-with-helm.md
chickenchickenlove: Thanks for your review.
And Good point. I agree with you 👍

jmazzitelli (changes requested):
Check the netlify page - https://deploy-preview-765--kiali.netlify.app/docs/installation/installation-guide/install-with-helm/#install-with-operator

As I suspected ("See my suggestion (make sure that link works :)") , the link doesn't work :) I think you need to prefix it with `/doc`.  So `/doc/configuration/authentication/`.

jmazzitelli (approved):

content/en/docs/Installation/installation-guide/install-with-helm.md
```diff
@@ -69,6 +69,7 @@ command:
 $ helm install \
     --set cr.create=true \
     --set cr.namespace=istio-system \
+    --set cr.auth.strategy="anonymous" \
```
jmazzitelli, line 72: This is my fault. I did not properly review this PR.

This is wrong and does nothing to affect the strategy. It should have been:

```
    --set cr.spec.auth.strategy="anonymous" \
```

(note the missing `.spec`)

content/en/docs/Installation/installation-guide/install-with-helm.md
```diff
@@ -80,7 +81,10 @@ create the `kiali-operator` namespace (if needed), and deploy the Kiali
 operator on it.  The `--set cr.create=true` and `--set
 cr.namespace=istio-system` flags instructs to create a Kiali CR in the
 `istio-system` namespace. Since the Kiali CR is created in advance, as soon as
-the Kiali operator starts, it will process it to deploy Kiali.
+the Kiali operator starts, it will process it to deploy Kiali. After Kiali has started,
+you can access Kiali UI through 'http://localhost:20001' by executing 
+`kubectl port-forward service/kiali -n istio-system 20001:20001`
+because of `--set cr.auth.strategy="anonymous"`. But realize that anonymous mode will allow anyone to be able to see and use Kiali. If you wish to require users to authenticate themselves by logging into Kiali, use one of the other [auth strategies](/docs/configuration/authentication/).
```
jmazzitelli, line 87: and here, too - should be "cr.spec.auth.strategy"

chickenchickenlove: @jmazzitelli sorry to bother you.
you right, i missed it. I'm really sorry.  
From [this](https://github.com/kiali/helm-charts/blob/fb5c5f760c9b8d10890620fa28f63924028bd228/kiali-operator/templates/kiali-cr.yaml#L19-L21), as you said, `cr.auth.strategy="anonymous"` should be change to `cr.spec.auth.strategy="anonymous"`, i think.

May i create new PR to fix it?

jmazzitelli: I already corrected it. It's live on the site: https://kiali.io/docs/installation/installation-guide/install-with-helm/#install-with-operator

chickenchickenlove: I see.
Thanks for your help. 🙇‍♂️

PR #797, opened 2024-06-25
Mesh page docs

ferhoyos:
I have added references to the mesh page in the following pages of the kiali.io site:
- Configuration / Istio environment
- Features / Istio status
- Features / Multi-cluster


jshaughn (changes requested):
All of the new content is great, no changes requested.  Thanks for the numerous screenshots.

One thing needs to be omitted: See https://deploy-preview-797--kiali.netlify.app/docs/features/multi-cluster/#list-view-mesh-discovery

ferhoyos:
Regarding the screenshots, I think we should replace a bunch of them across the site as they are quite outdated. Given the significant effort required to review the entire site, perhaps a good time to do this would be after upgrading to the new theme with PF6.

jshaughn (approved):
Looks great, thanks!

PR #798, opened 2024-06-27
Clarify KIA0106 to indicate that a pod must reference the account.

### LINKED ISSUE #7481: KIA0106 False Positive - Unable to Find Service Accounts
### Describe the bug

Given an `AuthorizationPolicy` in the `istio-system` namespace that specifies use of a `ServiceAccount` in a different namespace, we're receiving KIA0106 (service account not found for this principal) false positives since Kiali apparently can't find the service account that is there.

### Expected Behavior

The service account noted in the principal should be found.

### What are the steps to reproduce this bug?

`AuthorizationPolicy` like this:

```yaml
apiVersion: security.istio.io/v1
kind: AuthorizationPolicy
metadata:
  name: admin-app-policy
  namespace: istio-system
spec:
  action: DENY
  rules:
  - from:
    - source:
        principals:
        - cluster.local/ns/other-namespace/sa/admin-app
    to:
    - operation:
        notHosts:
        - my-host.my-namespace.svc.cluster.local:8080
```

And the noted service account exists (`cluster.local/ns/other-namespace/sa/admin-app`).

When you look at the policy in Kiali you get a KIA0106 error saying `cluster.local/ns/other-namespace/sa/admin-app` can't be found.

The policy works correctly in Istio - Istio can find the service account, and `kubectl get serviceaccount/admin-app -n other-namespace` does correctly return the service account.

### Environment

- **Kiali version:** v1.86.0 (5938c917c8e720450304d0f4041ec179f73a2fcc)
- **Istio version:** 1.20.3
- **Kubernetes impl:** Azure Kubernetes Service
- **Kubernetes version:** 1.28.5
- **Other notable environmental factors:** Deployed in read-only mode via Kiali Operator. Accessible namespaces set to `**`.

tillig:
Fixes kiali/kiali#7481.

Clarifies that an error on KIA0106 doesn't necessarily mean the service account doesn't exist - it may also mean that no pod in the system references the service account.

Corrected some grammar/punctuation issues on KIA0106 and related `AuthorizationPolicy` / Source rules to read in a slightly more natural fashion.

Removed some trailing whitespace from lines.

### DIFF
--- a/content/en/docs/Features/validations.md
+++ b/content/en/docs/Features/validations.md
@@ -87,10 +87,10 @@ Either remove the host from the list, correct if there is any typo or deploy a n
 
 ### KIA0105 - This field requires mTLS to be enabled
 
-AuthorizationPolicy has a Source field, where specifies the source identities of a request.
-In a Source field it accepts the principals and the namespaces, which requires mTLS enabled.
+AuthorizationPolicy has a Source field which specifies the source identities of a request.
+The Source field accepts principals and namespaces which require mTLS be enabled.
 
-A validation Error message on a principals or namespaces fields means, that mTLS is not enabled.
+A validation Error message on a principals or namespaces fields means that mTLS is not enabled.
 
 This validation appears only when autoMtls is disabled.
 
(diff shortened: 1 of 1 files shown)


hhovsepy:
Thank you @tillig for the contribution.
LGTM, waiting for final approval from @jshaughn

jshaughn (approved):
Yes, more clear, thanks!

PR #807, opened 2024-08-03
docs for the new discovery selector feature

jmazzitelli:
part of: kiali/kiali#7546
part of KEP: https://github.com/kiali/kiali/blob/master/design/KEPS/namespace-discovery/proposal.md

server PR: https://github.com/kiali/kiali/pull/7592
operator PR: https://github.com/kiali/kiali-operator/pull/797
helm chart PR: kiali/helm-charts#274
docs PR: https://github.com/kiali/kiali.io/pull/807

Netlify links:
* https://deploy-preview-807--kiali.netlify.app/docs/faq/installation/#operator-fails-due-to-cannot-list-resource-clusterroles-error
* https://deploy-preview-807--kiali.netlify.app/docs/faq/general/#why-isnt-my-namespace-in-the-namespace-selection-dropdown
* https://deploy-preview-807--kiali.netlify.app/docs/installation/installation-guide/example-install/
* https://deploy-preview-807--kiali.netlify.app/docs/configuration/namespace-management/


jmazzitelli:
In `content/en/docs/FAQ/installation.md` around line 10, we should ultimately change the link so it points to the latest values.yaml file (it is currently pointing to the 1.88.0 version of the file, but that's going to change after we merge all this stuff in - that version still references accessible_namespaces which it no longer will once we merge this stuff in).

I'm referring to this: `if you did so with the value [clusterRoleCreator](https://github.com/kiali/helm-charts/blob/v1.88.0/kiali-operator/values.yaml#L41-L45) set to false...`

content/en/docs/Configuration/namespace-management.md
```diff
@@ -5,188 +5,201 @@ description: "Configuring the namespaces accessible and visible to Kiali."
...
+It is possible to restrict Kiali so that it can only access a specific set of namespaces by providing [discovery selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements) that match those namespaces. Note that Kiali will not use [Istio's discovery selectors](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig); if Istio has been configured with its own discovery selectors, you will likely want to configure Kiali with the same list of discovery selectors.
+
+{{% alert color="info" %}}
+This documentation makes a distinction between a namespace that is _accessible_ verses _visible_. Kiali can be given access to all namespaces in the cluster, or it can be given restricted permissions so that only certain namespaces are accessible to it. Of the accessible namespaces that Kiali is permitted to see, it can make some or all of them visible to end users.
+{{% /alert %}}
 
 {{% alert color="warning" %}}
-As of Kiali 1.67, the following settings are deprecated:
+As of Kiali 1.89, the following settings are no longer supported:
```
nrfox: 2.0?

jmazzitelli: that's weird.. I fixed that yesterday.. lemme see where that went :)

jmazzitelli: ah! its in my other branch where I'm getting ready to add back the istio discovery selectors https://github.com/jmazzitelli/kiali.io/commit/ff3ea7bea8886e4dc7bc539ea1ae6674f69735d8#diff-18799846877673dfb4a2b103af33b69fece2ba435d3c326fa4c55ac7bc480920R17

Let me put that in this PR

jmazzitelli: done.

jmazzitelli:
we can ignore the validate error. that link (https://istio.io/v1.10/docs/setup/additional-setup/cni/#identifying-pods-requiring-traffic-redirection) really exists, dunno why the validator is getting a timeout.

content/en/docs/Configuration/namespace-management.md
```diff
@@ -5,188 +5,201 @@ description: "Configuring the namespaces accessible and visible to Kiali."
...
+It is possible to restrict Kiali so that it can only access a specific set of namespaces by providing [discovery selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements) that match those namespaces. Note that Kiali will not use [Istio's discovery selectors](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig); if Istio has been configured with its own discovery selectors, you will likely want to configure Kiali with the same list of discovery selectors.
+
+{{% alert color="info" %}}
+This documentation makes a distinction between a namespace that is _accessible_ verses _visible_. Kiali can be given access to all namespaces in the cluster, or it can be given restricted permissions so that only certain namespaces are accessible to it. Of the accessible namespaces that Kiali is permitted to see, it can make some or all of them visible to end users.
```
jshaughn:
```suggestion
This documentation makes a distinction between an _accessible_ and _visible_ namespace. Kiali can be given access to all namespaces in the cluster, or it can be given restricted permissions so that only certain namespaces are accessible to it. Of the accessible namespaces that Kiali is permitted to see, it can make some or all of them visible to end users.
```

I fixed the grammatical error, "verses" is like verses of the Bible.  I think you wanted "versus", but even that's not quite right.  But more than that, I don't totally understand this blurb.  It would be better if you describe exactly the difference between an _accessible_ namespace and a _visible_ namespace.  Even I don't know what you're saying here.

jshaughn: Meaning, what exactly does it mean for Kiali to have access?  What level of access, etc.

content/en/docs/Configuration/namespace-management.md
```diff
@@ -5,188 +5,201 @@ description: "Configuring the namespaces accessible and visible to Kiali."
...
-visible to the Kiali operator. If left unset, the Kiali server will be given access to all namespaces in
-the cluster.
+With cluster-wide access disabled, the Kiali Server must be told what namespaces are accessible to it.  These accessible namespaces are defined by a list of discovery selectors that match namespaces.
 
-The list of accessible namespaces is specified in the Kiali CR via the
-`accessible_namespaces` setting, under the main `deployment` section. As an
-example, if Kiali is to be installed in the `istio-system` namespace, and is
-expected to monitor all namespaces prefixed with `mycorp_`, the setting would
-be:
+The list of accessible namespaces is specified in the Kiali CR via the `deployment.discovery_selectors.default` setting. As an example, if Kiali is to be installed in the `istio-system` namespace, and is expected to monitor all namespaces with the label "my-mesh", the setting would be:
```
jshaughn, line 41: Does Kiali's install namespace need to be labeled?

content/en/docs/Configuration/namespace-management.md
```diff
@@ -5,188 +5,201 @@ description: "Configuring the namespaces accessible and visible to Kiali."
...
+    discovery_selectors:
+      default:
+      - matchExpression:
+        - key: my-mesh
+          operator: Exists
 ```
 
-When you set `accessible_namespaces` to a list of namespaces as the example above illustrates, by default the operator will create a Role for each accessible namespace. Each Role will be assigned to the Kiali Service Account thus providing Kiali permissions to those namespaces. To change that default behavior, `deployment.cluster_wide_access` can be set to `true`. This will tell the operator not to create these Roles and just create a single ClusterRole and assign it to the Kiali Server (see below).
+When `cluster_wide_access` is set to `false`, the Kiali Operator will examine the `default` selectors under `spec.deployment.discovery_selectors`, as the example above illustrates. The Kiali Operator will then attempt to find all the namespaces that match the discovery selectors. For each namespace that matches the discovery selectors, the Kiali Operator will create a Role and assign that Role to the Kiali Service Account thus giving Kiali access to those namespaces. These namespaces are therefore called the "accessible namespaces".
```
jshaughn:
```suggestion
When `cluster_wide_access` is set to `false`, the Kiali Operator will examine the `default` selectors under `spec.deployment.discovery_selectors`, as the example above illustrates. The Kiali Operator will then attempt to find all of the namespaces that match the discovery selectors. For each namespace that matches the discovery selectors, the Kiali Operator will create a Role and assign that Role to the Kiali Service Account thus giving Kiali access to those namespaces. These namespaces are therefore called the "accessible namespaces".
```

content/en/docs/Configuration/namespace-management.md
```diff
@@ -5,188 +5,201 @@ description: "Configuring the namespaces accessible and visible to Kiali."
...
 {{% /alert %}}
 
-When installing multiple Kiali instances into a single cluster,
-`accessible_namespaces` must be mutually exclusive. In other words, a namespace
-set must be matched by only one Kiali CR. Regular expressions must not have
-overlapping patterns.
-
 {{% alert color="warning" %}}
-A cluster can have at most one Kiali instance with `accessible_namespaces` unset.
+When installing multiple Kiali instances into a single cluster, `deployment.discovery_selectors.default` must be mutually exclusive. In other words, a namespace must be matched by discovery selectors defined in a single Kiali CR.
```
jshaughn:
```suggestion
When installing multiple Kiali instances into a single cluster, `deployment.discovery_selectors.default` must be mutually exclusive. In other words, a namespace must be matched by the discovery selectors defined by one and only one Kiali CR on the cluster.
```

jshaughn (commented):
I got to line 82 but I have to step out, here are some initial comments.  I'll finish later...

content/en/docs/Configuration/namespace-management.md
```diff
@@ -5,188 +5,201 @@ description: "Configuring the namespaces accessible and visible to Kiali."
...
 
-In Istio's [MeshConfig](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig), you can provide a list of discovery selectors that Istio will consider when processing namespaces. The Kiali server will also utilize these same discovery selectors when determining which namespaces to make available to users. Therefore, these discovery selectors work in conjunction with the `accessible_namespaces` setting. The `accessible_namespaces` setting dictates which namespaces are _accessible_ to the Kiali Server, and hence to users of Kiali. The Istio discovery selectors tell Kiali to make _available_ to users those namespaces within the `accessible_namespaces` list. So discovery selectors are a subset of `accessible_namespaces`.
+In Istio's [MeshConfig](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig), a list of discovery selectors can be configured. These Istio discovery selectors define the namespaces that Istio will consider "in the mesh." (see [this blog post](https://istio.io/v1.13/blog/2021/discovery-selectors/) for details). These Istio discovery selectors are utilized only by Istio; they will be ignored by Kiali.
```
jshaughn:
```suggestion
In Istio's [MeshConfig](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig), a list of discovery selectors can be configured. These Istio discovery selectors define the namespaces that Istio will consider "in the mesh" (see [this blog post](https://istio.io/v1.13/blog/2021/discovery-selectors/) for details). These Istio discovery selectors are utilized only by Istio; they will be ignored by Kiali.
```

content/en/docs/Configuration/namespace-management.md
```diff
@@ -5,188 +5,201 @@ description: "Configuring the namespaces accessible and visible to Kiali."
...
+Note that the discovery selectors are evaluated by the Kiali Operator at install time when deciding which namespaces should be accessible (and thus which Roles to create). Namespaces that do not exist at the time of install will not be accessible to Kiali until the operator has a chance to reconcile the Kiali CR. There are several ways in which the operator can be told to reconcile a Kiali CR in order to determine the new set of accessible namespaces.
 
+1. You can ask that the Kiali Operator periodically reconcile the Kiali CR on a fixed schedule. See the [Ansible Operator SDK documentation describing the reconcile-period annotation](https://sdk.operatorframework.io/docs/building-operators/ansible/reference/advanced_options/#ansiblesdkoperatorframeworkioreconcile-period-custom-resource-annotation). In short, you can have the Kiali Operator periodically reconcile a Kiali CR by setting the `ansible.sdk.operatorframework.io/reconcile-period` annotation on the Kiali CR. For example, to reconcile this Kiail CR every 60 seconds:
```
jshaughn:
```suggestion
1. You can ask that the Kiali Operator to periodically reconcile the Kiali CR on a fixed schedule. See the [Ansible Operator SDK documentation describing the reconcile-period annotation](https://sdk.operatorframework.io/docs/building-operators/ansible/reference/advanced_options/#ansiblesdkoperatorframeworkioreconcile-period-custom-resource-annotation). In short, you can have the Kiali Operator periodically reconcile a Kiali CR by setting the `ansible.sdk.operatorframework.io/reconcile-period` annotation on the Kiali CR. For example, to reconcile this Kiail CR every 60 seconds:
```

content/en/docs/Configuration/namespace-management.md
```diff
@@ -5,188 +5,201 @@ description: "Configuring the namespaces accessible and visible to Kiali."
...
+3. Similar to the above, touching any annotation on the Kiali CR will also trigger a reconciliation. One suggestion is to dedicate an annotation whose purpose is solely to trigger operator reconcilations. For example, add or modify the "trigger-reconcile" annotation on the Kiali CR to trigger the operator to run a reconcilation on that Kiali CR:
+```sh
+kubectl annotate kiali my-kiali-cr --namespace istio-system --overwrite trigger-reconcile="$(date)"
 ```
+4. The Kiali Operator can be enabled to watch for namespaces getting created in the cluster. When new namespaces are created, the Kiali Operator will detect this and will then attempt to reconcile all Kiali CRs in the cluster. To enable operator namespace watching, see the [FAQ]({{< ref "/docs/faq/installation" >}}) describing the operator WATCHES_FILE environment variable. Note that on clusters with large numbers of namespaces that get created, enabling this namespace watching feature can cause the operator to consume alot of CPU, so you may not wish to use this method.
```
jshaughn:
```suggestion
4. The Kiali Operator can be enabled to watch for namespaces getting created in the cluster. When new namespaces are created, the Kiali Operator will detect this and will then attempt to reconcile all Kiali CRs in the cluster. To enable operator namespace watching, see the [FAQ]({{< ref "/docs/faq/installation" >}}) describing the operator WATCHES_FILE environment variable. Note that on clusters with large numbers of namespaces that get created, enabling this namespace watching feature can cause the operator to consume a lot of CPU, so you may not wish to use this method.
```

jshaughn (changes requested):
It's good, a few minor changes.  I hope this approach is actually better than what we do currently.

content/en/docs/Configuration/namespace-management.md
jmazzitelli: `This documentation makes a distinction between an _accessible_ and a _visible_ namespace. The Kiali Server will be given permission to see either (a) all namespaces in the cluster, or (b) only certain namespaces in the cluster. These are the "accessible namespaces." This means the Kiali Server will be able to query for and access resources within those accessible namespaces, but be forbidden from seeing all the other namespaces and their resources. Of these accessible namespaces that Kiali is permitted to see, it can make some or all of them visible to end users such that the end users will be able to see those namespaces and their resources in the Kiali UI.`

This is important because its the distinction between "cluster_wide_access" values of "true" and "false" which the documentation will later talk about. e.g. when you have CWA=true, the Kiali Server itself can access all namespaces in the cluster (they are "accessible") but you could still set discovery selectors to limit the visibility to only some namespaces to end users. But I didn't want to get into the details here because its all described later in the docs - but I did want to point out the distinction between "accessible" and "visible" because its important later.

Why do we do this? It's the whole cluster-scoped versus namespace-scoped caches (have a cluster-scoped cache for efficiency, have namespace-scoped caches for higher security with the downside of having many more k8s informers/watchers in the server).

content/en/docs/Configuration/namespace-management.md
jmazzitelli: no, because later the docs say (line 57):

`The Kiali Operator will always give the Kiali Server access to the namespace where the Kiali Server is installed and to the Istio control plane namespace (which may be the same namespace), whether those namespaces match a discovery selector or not.`

content/en/docs/Configuration/namespace-management.md
jmazzitelli: done

content/en/docs/Configuration/namespace-management.md
jmazzitelli: done

content/en/docs/Configuration/namespace-management.md
jmazzitelli: done

content/en/docs/Configuration/namespace-management.md
jmazzitelli: done

content/en/docs/Configuration/namespace-management.md
jmazzitelli: done

content/en/docs/Configuration/namespace-management.md
jshaughn: OK, how about this:
This documentation makes a distinction between _accessible_ and _visible_ namespaces. The Kiali Server will be given permission to access either (a) all, or (b) a configured subset, of cluster namespaces. The Kiali Server will only be aware of, query for, and access resources within these accessible namespaces. The set of namespaces visible to an end user, via the Kiali UI, will be a subset of the accessible namespaces.  In other words, the namespaces visible to a user may be all, or just some of the namespaces accessible to the Kiali Server..

jmazzitelli: see latest commit

jshaughn (approved):
LGTM!

PR #815, opened 2024-09-03
FAQ on how to declare credentials within secrets when using server helm chart

jmazzitelli:
Netlify preview: https://deploy-preview-815--kiali.netlify.app/docs/faq/installation/#how-can-i-use-a-secret-to-pass-external-service-credentials-to-the-kiali-server

I did a quick smoke test and this appears to work. Probably need some more testing to confirm these instructions actually do work.


jmazzitelli:
I did more testing and it appears the documentation is all correct.

jshaughn (approved):
Seems clear.

PR #821, opened 2024-09-26
Remove unnecessary documentation about secret permissions management

leandroberetta:
This documentation is not required anymore as the certificate is not coming from a Secret anymore.


jmazzitelli (approved):
lgtm

PR #823, opened 2024-09-30
Relnotes v2.0

jshaughn:


jmazzitelli (approved):
some TBDs in here which we will fill out later - this is just getting committed to staging for now, which eventually will be the 2.0 release notes.

PR #831, opened 2024-10-31
talk about using the operator to create the remote cluster resources

jmazzitelli:
part of https://github.com/kiali/kiali/issues/7861

netlify: https://deploy-preview-831--kiali.netlify.app/docs/configuration/multi-cluster/#setup

### DIFF
--- a/content/en/docs/Configuration/multi-cluster.md
+++ b/content/en/docs/Configuration/multi-cluster.md
@@ -48,37 +48,6 @@ Use the option `--help` for additional details on using the script to create and
 
 3. **Configure Kiali.** The Kiali CR provides configuration settings that enable the Kiali Server to use remote cluster secrets in order to access remote clusters. By default, the Kiali Operator will [auto-detect](/docs/configuration/kialis.kiali.io/#.spec.clustering.autodetect_secrets) any remote cluster secret that has a label `kiali.io/multiCluster=true` and is found in the Kiali deployment namespace. The secrets created by the `kiali-prepare-remote-cluster.sh` script will be created that way and thus can be auto-detected. Alternatively, in the Kiali CR you can [explicitly specify each remote cluster secret](/docs/configuration/kialis.kiali.io/#.spec.clustering.clusters) rather than rely on auto-discovery. Given the remote cluster secrets it knows about (either through auto-discovery or through explicit configuration) the Operator will mount the remote cluster secrets into the Kiali Server pod effectively putting Kiali in "multi-cluster" mode. Kiali will begin using those credentials to communicate with the other clusters in the mesh.
 
-3. Optional - **Configure tracing with cluster ID.** By default, traces do not include their cluster name in the trace tags; however, this can be added using the Istio telemetry API.
-
-```
-kubectl apply -f - <<EOF
-apiVersion: telemetry.istio.io/v1alpha1
-kind: Telemetry
-metadata:
-  name: mesh-default
-  namespace: istio-system
-spec:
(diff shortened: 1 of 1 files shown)


content/en/docs/Configuration/multi-cluster.md
```diff
@@ -17,33 +17,38 @@ Before proceeding with the setup, ensure you meet the requirements.
...
 
-3. Optional - **Configure tracing with cluster ID.** By default, traces do not include their cluster name in the trace tags however this can be added using the istio telemetry API.
+3. Optional - **Configure tracing with cluster ID.** By default, traces do not include their cluster name in the trace tags; however, this can be added using the Istio telemetry API.
```
nrfox: This is actually no longer true as of some recent Istio verison. Will you remove this section?

content/en/docs/Configuration/multi-cluster.md
```diff
@@ -17,33 +17,38 @@ Before proceeding with the setup, ensure you meet the requirements.
...
 
-   - Create a Kiali Service Account in the remote cluster.
-   - Create a role/role-binding for this service account in the remote cluster.
-   - Create a kubeconfig file and save this as a secret in the namespace where Kiali is deployed.
+2. **Create a remote cluster secret.** In order for Kiali to access a remote cluster, you must provide a kubeconfig to Kiali via a Kubernetes secret. This requires you to obtain a token for the remote cluster's SA created in step 1. It is up to you how you want to create and manage this token, however, you can use the [kiali-prepare-remote-cluster.sh script](https://github.com/kiali/kiali/blob/master/hack/istio/multicluster/kiali-prepare-remote-cluster.sh) to simplify this process for you.
```
nrfox: I know the script can run `helm template | kubectl apply` but I think we should recommend that people only use the script for creating the remote secret. For creating the remote resources, users can run `helm install kiali-server --set remote-resources-only=true` on their own or if they are using the operator then creating the Kiali CR with that option.

content/en/docs/Configuration/multi-cluster.md
jmazzitelli: I was just changing the punctuation and capitalization of an already existing section. If this is no longer accurate, I can remove this line entirely.

jmazzitelli: this section has been removed

content/en/docs/Configuration/multi-cluster.md
jmazzitelli: Right now, the default for the script is [to create both the secret and remote resources](https://github.com/kiali/kiali/blob/v2.0.0/hack/istio/multicluster/kiali-prepare-remote-cluster.sh#L30-L31). Now that we have support for remote resources in operator and helm charts, and now that we'll recommend using the script only to create the secret, maybe we should make the default behavior to only create the secret? So we'd change it to:
```bash
DEFAULT_PROCESS_KIALI_SECRET="true"
DEFAULT_PROCESS_REMOTE_RESOURCES="false"
```

The user could use the script to create both, they would just have to pass in [`--process-remote-resources=true` ](https://github.com/kiali/kiali/blob/v2.0.0/hack/istio/multicluster/kiali-prepare-remote-cluster.sh#L474) explicitly. But this way for the documented/recommended way to use the script, they wouldn't have to pass either of the --process-xyz options, the defaults would just create the secret.

nrfox: I'm good with either way.

nrfox (approved):

PR #835, opened 2024-11-18
document the need for different instance names if CWA=true

### LINKED ISSUE #7922: allow multiple Kiali Servers in the same cluster each have cluster-wide-access enabled
The docs ( https://kiali.io/docs/configuration/namespace-management/#cluster-wide-access-mode ) say this:

```
For clusters that have multiple Kiali Servers installed, only one of them
may be installed with deployment.cluster_wide_access set to true.
```

So it claims you cannot have 2 or more Kiali Servers installed in the cluster all with CWA=true.

The reason for this restriction is because the operator creates a cluster role/rolebinding when CWA=true and if multiple Kialis are installed with CWA=true there is a conflict (e.g. if one of them is deleted, the cluster role/binding gets deleted and break all other Kialis with CWA=true; similar with adding new Kialis and it overwrites the cluster role binding -- see comments below for a demonstration of this).

UPDATE: But after some more testing, I realized there already is support for this. You can have multiple Kialis with CWA=true today - you just need to provide unique instance names to them (so their `spec.deployment.instance_name` values must be different across all Kiali servers that have CWA=true).

We need to update the docs to mention this.

We should also update the operator to prohibit installing two Kiali Servers with the same instance_name if they both want CWA=true.

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7922

netlify: https://deploy-preview-835--kiali.netlify.app/docs/configuration/namespace-management/#cluster-wide-access-mode

### DIFF
--- a/content/en/docs/Configuration/namespace-management.md
+++ b/content/en/docs/Configuration/namespace-management.md
@@ -26,7 +26,9 @@ As of Kiali 2.0, the following settings are no longer supported:
 
 By default, the Kiali Server is given cluster-wide access to all namespaces on the local cluster. This is controlled by the Kiali CR setting `deployment.cluster_wide_access`, which has a default value of `true` when not specified.
 
-For clusters that have multiple Kiali Servers installed, only one of them may be installed with `deployment.cluster_wide_access` set to `true`.
+{{% alert color="info" %}}
+You cannot have multiple Kiali Servers with cluster wide access that have identical instance names. If you wish to install multiple Kiali Servers with cluster wide access enabled, they each must have unique `deployment.instance_name` values.
+{{% /alert %}}
 
 In order to restrict the Kiali Server so that it only has access to certain namespaces on the local cluster, it must first have its cluster-wide access disabled. You do this by setting `deployment.cluster_wide_access` to `false` in the Kiali CR.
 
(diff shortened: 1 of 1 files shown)


content/en/docs/Configuration/namespace-management.md
```diff
@@ -26,7 +26,9 @@ As of Kiali 2.0, the following settings are no longer supported:
 
 By default, the Kiali Server is given cluster-wide access to all namespaces on the local cluster. This is controlled by the Kiali CR setting `deployment.cluster_wide_access`, which has a default value of `true` when not specified.
 
-For clusters that have multiple Kiali Servers installed, only one of them may be installed with `deployment.cluster_wide_access` set to `true`.
+{{% alert color="info" %}}
+You cannot have multiple Kiali Servers with cluster wide access that have identical instance names. If you wish to install multiple Kiali Servers with cluster wide access enabled, they each must have unique `deployment.instance_name` values.
```
jshaughn:
```suggestion
You cannot have multiple Kiali Servers with both cluster-wide access and identical instance names. If you wish to install multiple Kiali Servers with cluster-wide access enabled, each must have a unique `deployment.instance_name` value.
```

jshaughn (changes requested):
Please accept my humble update.

jshaughn (approved):
fabulous

PR #838, opened 2024-12-02
Add docs for configuring custom CA on openshift

nrfox:
Adds documentation for adding custom CA certs to Kiali for remote oAuth servers using the `openshift` strategy.

https://deploy-preview-838--kiali.netlify.app/docs/configuration/authentication/openshift/#multi-cluster---using-an-internal-or-self-signed-certificate

### DIFF
--- a/content/en/docs/Configuration/authentication/openshift.md
+++ b/content/en/docs/Configuration/authentication/openshift.md
@@ -56,7 +56,7 @@ After restarting the Kiali pod, Kiali will trust this root certificate for all H
 #### Insecure setting
 
 {{% alert color="warning" %}}
-You should use this setting for testing purposes only.
+You should only use this setting for testing and not in a production environment.
 {{% /alert %}}
 
 You can disable certificate validation between Kiali and the remote OAuth server(s) by setting `insecure_skip_verify_tls` to `true` in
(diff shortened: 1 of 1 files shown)


content/en/docs/Configuration/authentication/openshift.md
```diff
@@ -36,3 +36,35 @@ Kiali as a client for the most common use-cases. The `openshift` strategy does h
 configuration settings that most people will never need but are available in case you have
 a situation where the customization is needed. See the Kiali CR Reference page for the
 documentation on those settings.
+
+### Multi-Cluster - Using an internal or self-signed certificate
+
+If you have a multi-cluster Kiali deployment and the OAuth server is configured with an external IdP that uses an internal or self-signed certificate, you can configure Kiali to trust the server's certificate by creating a ConfigMap named `kiali-oauth-cabundle` containing the CA certificate bundle for the server under the `oauth-server-ca.crt` key:
+
```
jmazzitelli, line 43: The Kiali deployment will project a ConfigMap whose name is "-oauth-cabundle" prefixed with the kiali instance name [as seen here in the PR](https://github.com/kiali/kiali-operator/pull/853/files#diff-fc1479bd47964311cee57d193b73019bcb1090757edc493dac0c3b122809cfc0R179) (so by default it will look for `kiali-oauth-bundle` because the default instance name is `kiali`). But if Kiali is configured with a different instance name (which is possible), the user needs to know to create the configmap with the proper name. I suggest an info block here that says something like this:

```suggestion

{{% alert color="info" %}}
Note that if you are deploying Kiali with `spec.deployment.instance_name` set to a value that is different than the default of `kiali`, your ConfigMap name needs to be that instance name appended with "-oauth-bundle". For example, if your instance name is "myserver" then the name of the ConfigMap must be `myserver-oauth-cabundle`.
{{% /alert %}}

```

jmazzitelli (changes requested):
LGTM except for the following...

jmazzitelli (approved):
lgtm

PR #842, opened 2024-12-20
document that Kiali is not affected by CVE-2024-2961

jmazzitelli:
We've had several people report this CVE against Kiali Server, but Kiali Server is not affected. We need to document this.

Netlify: https://deploy-preview-842--kiali.netlify.app/news/security-bulletins/


jshaughn (approved):
LG and gives us something to point to.

PR #849, opened 2025-01-07
Update compatibility approach to include all supported istio versions at kiali-release-time

jshaughn:
part-of https://github.com/kiali/kiali/issues/7932

https://deploy-preview-849--kiali.netlify.app/docs/installation/installation-guide/prerequisites/


jmazzitelli (approved):
lgtm

data/compatibility/istio.yaml
nrfox, line 1: Following the pattern a few versions out, the support range look something like this, right?:
```
1.26: 2.12
1.25: 2.8 - 2.12
1.24: 2.4 - 2.12
```

jshaughn: That's right, the idea being that users can feel more confident updating to newer versions, and also know that that version can help them with a revision-based update. So I may be running 1.24 with Kiali 2.4.  But I want to update to 1.26.  Before I do I can update Kiali to 2.12, which should be compatible with both versions.

nrfox: Ok the existing matrix looks a little funny with `    kialiVersions: "1.87 to 2.0, 2.4 "` but I think that's just because 2.4 is the first version that is officially doing -2 version support?

jshaughn: that's right. I would expect that we could say 1.87 to 2.4, but that's not really tested, so I went with this unusual one until it all evens out.

nrfox (approved):

PR #861, opened 2025-02-17
Add mesh page to the OSSMC user guide

ferhoyos:
This PR adds the mesh page description in the OSSMC user guide


jmazzitelli (approved):
very nice!

PR #897, opened 2025-07-24
External Kiali Documentation

### LINKED ISSUE #8581: [kiali.io] external deployment model
Document on kiali.io the new external deployment model.

jshaughn:
Closes https://github.com/kiali/kiali/issues/8581

- add initial doco for external deployments
  - create a Configuration/Multi-cluster directory
    - push down the existing content
    - add an external deployment  section
- remove instances of spec.istio_namespace from the doco
- add some rel-note info about the newly ignored CR fields
- add a lame drawing

https://deploy-preview-897--kiali.netlify.app/docs/configuration/multi-cluster/external/


content/en/news/release-notes.md
```diff
@@ -6,6 +6,14 @@ weight: 1
 
 For additional information check our [sprint demo videos](https://www.youtube.com/channel/UCcm2NzDN_UCZKk2yYmOpc5w) and [blogs](https://medium.com/kialiproject).
 
+Upgrade Change Notes:
+
+The following fields are no longer used by the Kiali CR and may be safely removed, if set. Leaving them in an existing CR will not be a problem, but they are unused.
```
jmazzitelli: This likely needs to be re-worded if we get the CRD Schema merged along with this stuff. If that happens, then it is no longer true that leaving them in will not be a problem. It will be a problem - the CR will fail schema validation and the Kiali CR will be rejected.

We should say something like, "These are no longer used and must be removed."

content/en/docs/Configuration/multi-cluster/external.md
```diff
@@ -0,0 +1,33 @@
...
+It is recommended, but not required, that joining Kiali on the "management" cluster, you also place your other observability tooling, like the metrics store. This will further reduce observability resources on the mesh cluster(s), and will likely reduce latency between Kiali and those data stores.
+
+The high-level deployment model looks like this:
+![Kiali multi-cluster](/images/documentation/configuration/external-kiali.png)
+
+### Configuration
+
+Configuring Kiali for the external deployment model has the same requirements needed for a co-located Kiali in a [multi-cluster installation]({{< relref "../Multi-cluster" >}}). Kiali still needs the necessary secrets for accessing the remote clusters. 
+
+Additionally, The configuration needs to indicate that Kiali will not be managing its home cluster. This is done in the Kiali CR by setting:
```
jmazzitelli:
```suggestion
Additionally, the configuration needs to indicate that Kiali will not be managing its home cluster. This is done in the Kiali CR by setting:
```

content/en/docs/Configuration/multi-cluster/external.md
```diff
@@ -0,0 +1,33 @@
+---
+title: "External Kiali"
+description: "Deploy Kiali on a Management Cluster."
+---
+
+Larger mesh deployments may desire to separate mesh operation from mesh observability. By deploying Kiali on a cluster away from the mesh, it allows for dedicated management and resource consumption.
+
+### Deployment Model
+
+This deployment model requires a minimum of two clusters. The "management" cluster is where Kiali will be deployed. The "mesh" cluster(s) will be where your service mesh is deployed. The mesh deployment will still conform to any of the Istio deployment models that Kiali already supports. The fundamental difference is that Kiali will not be co-located with an Istio control plane, but instead will reside away from the mesh. For multi-cluster mesh deployments, all of the same requirements apply, such as unified metrics and traces, etc.
```
jmazzitelli: We should introduce the term "home cluster" early on in these docs, since it is something users should know about. One suggestion is:

```suggestion
This deployment model requires a minimum of two clusters. The Kiali "home" cluster (where Kiali is deployed) will serve as the "management" cluster. The "mesh" cluster(s) will be where your service mesh is deployed. The mesh deployment will still conform to any of the Istio deployment models that Kiali already supports. The fundamental difference is that Kiali will not be co-located with an Istio control plane, but instead will reside away from the mesh. For multi-cluster mesh deployments, all of the same requirements apply, such as unified metrics and traces, etc.
```

content/en/docs/Configuration/multi-cluster/external.md
```diff
@@ -0,0 +1,33 @@
...
+Configuring Kiali for the external deployment model has the same requirements needed for a co-located Kiali in a [multi-cluster installation]({{< relref "../Multi-cluster" >}}). Kiali still needs the necessary secrets for accessing the remote clusters. 
+
+Additionally, The configuration needs to indicate that Kiali will not be managing its home cluster. This is done in the Kiali CR by setting:
+
+```
+clustering:
+  ignore_home_cluster: true
+```
+
+Kiali typically sets its home cluster name to the same cluster name set by the co-located Istio control plane. In an external deployment there is no co-located Istio control plane, and therefore the cluster name must also be set in the configuration. This should be the same name as reflected in the .kube config.
```
jmazzitelli: `This should be the same name as reflected in the .kube config.`

What .kube config? That isn't explained. And what is `.kube` referring to. There is no $HOME/.kube for the Kiali pod, so this is going to be confusing.

IIUC, this cluster name can be anything. It only "should be the same name" as the remote cluster secret that run-kiali.sh set up , which is usually the name of the cluster in the context's kubeconfig" - but of course users don't care about run-kiali :) So I don't think we should say anything about  .kube config. I think this value can be anything AS LONG AS it is not the same name as one of the remote clusters.

jmazzitelli (changes requested):

jmazzitelli (approved):

content/en/docs/Configuration/multi-cluster/external.md
nrfox, line 1: @jshaughn we need to mention that this is only supported with the anonymous auth strategy right now. Maybe it should even be labeled as "experimental" for the time being? At least until there is automated testing for it.

jshaughn: Will update as needed before the release.

PR #899, opened 2025-08-07
more explaination about why system ns are excluded by default

### LINKED ISSUE #8615: Kiali errorneously filters out some of my namespaces
### Describe the bug

This bug affects the namespace selector used across Kiali. Some namespaces are filtered out and can't be selected.

The bug is caused by this regex: https://github.com/kiali/kiali/blob/0ef9f4e83ca472ad463592126aaa5c2c06fb2470/istio/discovery_selectors.go#L16

I have three namespaces (`kube-prometheus-stack`, `kube-state-metrics` and `kiali-operator`) that are part of my service mesh but can't be selected in Kiali. 

When I select "All namespaces", these workloads become visible in the Traffic Graph. Regardless of the namespace setting, they don't show up in the Overview.

### Expected Behavior

These namespaces shouldn't be filtered out by default. I know that I can define a custom discoverySelector to override this behavior, but it took quite some digging (as you see, even in the Kiali source code) to find that out. If Kiali wants to be opinionated about which namespaces should be selectable, it should tell the user when it decides to not show something to him. At the very least, the documentation should be improved.

### What are the steps to reproduce this bug?

1. Have workloads in a namespace that matches the systemNamespaceRegex (e.g. something starting with `kube-`). Ensure they are part of the service mesh, and that their Envoy/Istio metrics are scraped by Prometheus.
2. Have a Kiali instance with cluster-wide access and no custom discoverySelectors
3. Open Kiali
4. You are now unable to select this namespace or workload in most Kiali views

### Environment
Kiali v2.13.0 (a7629a3bdd9a06160f7121da1aa700139718e865)
Istio v1.26.1
Prometheus v3.4.0
kube-state-metrics v2.16.0
Kubernetes version: v1.32.6 (AKS cluster)

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/8615

netlify: https://deploy-preview-899--kiali.netlify.app/docs/configuration/namespace-management/#discovery-selectors

### DIFF
--- a/content/en/docs/Configuration/namespace-management.md
+++ b/content/en/docs/Configuration/namespace-management.md
@@ -162,7 +162,7 @@ The `default` and `overrides` discovery selectors are processed in the same mann
 
 {{% alert color="info" %}}
 An empty list of discovery selectors has different semantics depending on the value of `deployment.cluster_wide_access`.
-* If `deployment.cluster_wide_access` is `true`, an empty list of discovery selectors means all namespaces will be visible except those that are considered system namespaces (these include namespaces whose names are prefixed with "kube-", "openshift" or "ibm" such as `kube-system`, `openshift-operators`, and `ibm-system`).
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
LG, better explanation of why it "is like it is".

PR #915, opened 2025-09-22
update relnotes with note about external_services.istio.registry

jshaughn:
Part of https://github.com/kiali/kiali/issues/8692

### DIFF
--- a/content/en/news/release-notes.md
+++ b/content/en/news/release-notes.md
@@ -6,6 +6,16 @@ weight: 1
 
 For additional information check our [sprint demo videos](https://www.youtube.com/channel/UCcm2NzDN_UCZKk2yYmOpc5w) and [blogs](https://medium.com/kialiproject).
 
+## 2.17.0
+
+!!! RELNOTES GO HERE !!!
+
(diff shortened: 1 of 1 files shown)


content/en/news/release-notes.md
```diff
@@ -6,6 +6,16 @@ weight: 1
 
 For additional information check our [sprint demo videos](https://www.youtube.com/channel/UCcm2NzDN_UCZKk2yYmOpc5w) and [blogs](https://medium.com/kialiproject).
 
+## 2.17.0
+
+!!! RELNOTES GO HERE !!!
+
+#### Upgrade Change Notes:
+
+**The following fields are no longer used by the Kiali CR and MUST be removed, if currently set.**
```
josunect, line 15: Should they said that is no longer supported? It was removed from the server as well

jshaughn: This is the same messaging we have used below. We're telling folks to remove it from the CR, so I think it's good enough.

josunect (approved):
lgtm!

PR #921, opened 2025-10-21
external kiali setup on openshift

jmazzitelli:
netlify: https://deploy-preview-921--kiali.netlify.app/docs/configuration/multi-cluster/external-openshift/


jshaughn:
Closing this. I don't think we want a tutorial-style page here. We may want to use this file for a blog instead.

PR #929, opened 2025-12-04
Update OSSMC compatibility table

ferhoyos:
Update OSSMC compatibility table:
- OSSMC versions 1.73-1.83 are not supported in OCP 4.19+
- New OSSMC version 2.20 will be only compatible for OCP 4.19+


jshaughn (approved):
Thanks!

PR #932, opened 2025-12-22
update release schedule to start again with January 26 release, and…

jshaughn:
Update release schedule to start again with January 26 release, and then again every three weeks.

Also, trim the Releases drop-down to recent and pinned releases


.github/workflows/release.yaml
```diff
@@ -50,8 +50,8 @@ jobs:
         cat <<-EOF > minor.py
         import datetime
 
-        # The base date can be any end of sprint from the past
-        base = int(datetime.datetime.strptime("24/04/2022", "%d/%m/%Y").timestamp())
+        # The base date is set so that releases occur every 3 weeks starting from January 26, 2026
+        base = int(datetime.datetime.strptime("19/01/2026", "%d/%m/%Y").timestamp())
```
ferhoyos: I think we should disable release for the moment because we don't know exactly what happens with previous dates (diff between now and base is negative).

According to AI, with this PR there will be a release in December 29th of 2025:
```
Let me calculate what happens on December 29, 2025 with that code:

  Base date: January 19, 2026
  "Now" date: December 29, 2025

  Since December 29, 2025 is before the base date, the diff will be negative:

  diff = Dec 29, 2025 - Jan 19, 2026 = -21 days (in seconds)

  days_elapsed = -21
  weeks_elapsed = int(-21 / 7) = -3
  weeks_mod3 = -3 % 3 = 0  (Python modulo returns 0 for -3 % 3)

  Result: weeks_mod3 = 0

  According to Kiali's release logic, weeks_mod3 == 0 means it's a minor release day.

  So yes, if this code is applied, December 29, 2025 would incorrectly trigger a release because:
  1. The negative diff produces weeks_mod3 = 0
  2. The schedule was intended to start January 26, 2026, but the math doesn't account for dates before the base

  This appears to be a bug - the code should probably check that diff >= 0 before determining it's a release day, or use a base date that's in the past relative to when the schedule should start.
```

jshaughn: Bah, nothing is easy. Good catch.  I updated the PR to ensure we don't launch before the base date. This should make it easier in the future to just update the base date, when needed.

jshaughn: I also fixed some broken links.

jshaughn: Also, have the link proofer ignore more codes, it's always failing, I think due to rate limiting.

ferhoyos: I like the validation, but I don't understand why we need to set the base date 1 week before the release date (set the base date in January 19th to start the release in January 26th). I think it is better to set the base date the same date of the first release, then next time you won't have to calculate the base date, just set the first release date. To achieve that I think you only have to validate `weeks_mod3` is 0 instead of 1.

.github/workflows/release.yaml
```diff
@@ -60,7 +60,11 @@ jobs:
         weeks_elapsed = int(days_elapsed / 7)
         weeks_mod3 = int(weeks_elapsed % 3)
 
-        print(weeks_mod3)
+        # Only release when weeks_elapsed >= 0 (past the base date) and weeks_mod3 == 1
+        if weeks_elapsed >= 0 and weeks_mod3 == 1:
+            print(1)
+        else:
+            print(0)
```
ferhoyos:
```suggestion
        print (weeks_elapsed >= 0 and weeks_mod3 == 0) # True in case of release
```

.github/workflows/release.yaml
ferhoyos:
```suggestion
        if [[ $DO_RELEASE == "True" ]]
```

.github/workflows/release.yaml
jshaughn: I had it that way and there seemed to be an issue (according to cursor). Anyway, sorry this turned into a time sync. If you are confident I'm fine with it. I'll apply your changes and let's be done with it.

.github/workflows/release.yaml
ferhoyos:
```suggestion
        base = int(datetime.datetime.strptime("26/01/2026", "%d/%m/%Y").timestamp())
```

.github/workflows/release.yaml
```diff
@@ -60,14 +60,15 @@ jobs:
         weeks_elapsed = int(days_elapsed / 7)
         weeks_mod3 = int(weeks_elapsed % 3)
 
-        print(weeks_mod3)
+        # Only release when weeks_elapsed >= 0 (past the base date) and weeks_mod3 == 1
```
ferhoyos:
```suggestion
        # Only release when weeks_elapsed >= 0 (past the base date) and weeks_mod3 == 0
```

.github/workflows/release.yaml
jshaughn: I committed your suggestions and Cursor agrees that the it is fine.

ferhoyos (approved):
Thanks for accepting the suggestion, I think now it is more clear and consistent. The rest of the changes look good to me

PR #938, opened 2026-01-19
Fix negative rounding issue in the release schedule

ferhoyos:
### Describe the change

The release schedule calculation in the CI workflow could incorrectly trigger releases 1-6 days before the base date (January 26, 2026) due to Python's int() function truncating toward zero instead of flooring.

**Example of the bug**
3 days before base date: diff = -259200 seconds
int(-3/7) = 0 (truncates toward zero)
Condition weeks_elapsed >= 0 evaluates to True -> unintended release

**Root Cause**
Using int(x / y) for division truncates toward zero, which behaves differently for negative numbers compared to floor division:
```
Expression	int() 	// (floor)
-3 / 7	     0	        -1
-6 / 7	     0          -1
```

**Solution**
Replace int(x / y) with floor division (//), which correctly rounds toward negative infinity

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -56,9 +56,10 @@ jobs:
 
         diff = now - base
 
-        days_elapsed = int(diff / (24*60*60))
-        weeks_elapsed = int(days_elapsed / 7)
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
LG

PR #944, opened 2026-02-03
OpenShift ACM docs

jmazzitelli:
Generated-by: Cursor

Netlify: https://deploy-preview-944--kiali.netlify.app/docs/configuration/multi-cluster/acm-observability/

This is based on how configuration is done via the hack script found here: https://github.com/kiali/kiali/pull/9176

### DIFF
--- a/content/en/docs/Configuration/multi-cluster/acm-observability.md
+++ b/content/en/docs/Configuration/multi-cluster/acm-observability.md
@@ -27,7 +27,7 @@ Kiali can query these aggregated metrics either through ACM's external Observato
 - **User Workload Monitoring (UWM)**: OpenShift's Prometheus for user workloads
 - **PodMonitor/ServiceMonitor**: Scrape Istio sidecar and control plane metrics
 - **Metrics Allowlist ConfigMaps**: Define which metrics ACM should collect
-- **Metrics Collector**: Runs on each managed cluster and pushes its Prometheus metrics to the hub cluster's Thanos every 5 minutes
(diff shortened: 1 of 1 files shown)


jmazzitelli:
One thing that is missing is documentation for Ambient metric collection. I need to research how those are different (names of the metrics, where they are scraped) and determine how this doc needs to be updated for that.

_UPDATE_: I added ambient information to this doc.

jmazzitelli:
This is ready for people to review.

jmazzitelli:
Note, these instructions will only work with Kiali v2.21+ (that's when we added the cert stuff).

jshaughn (commented):
Great content. Certainly not simple. I wonder whether we should move the whole thing under the "External Kiali" section? I think if you're considering ACM, and the fact that we would recommend locating Kiali on the hub, that it probably would not make sense to deploy Kiali in a co-located way.

jmazzitelli:
> I wonder whether we should move the whole thing under the "External Kiali" section?

We should probably look at this more holistically - because the External Kiali section has its own OpenShift sub-section:

image

So perhaps we should look at this Multi-cluster section as a whole and figure out what to do with OpenShift-specific stuff. FWIW: because ACM does not require an External Kiali setup, I don't think it should be in there. But if that's the way people are, in practice, always going to install it, then maybe.

jshaughn:
> So perhaps we should look at this Multi-cluster section as a whole...

I'm not sure one way is better than the other. For now, let's just get the content in. I'll approve and merge.

jshaughn (approved):
Great content. We may want to follow up with some re-org of the MC section, but lets get this in.

PR #946, opened 2026-02-10
Document use_waypoint_name setting

josunect:
https://deploy-preview-946--kiali.netlify.app/docs/features/ambient/#ambient-tracing

Ref. https://github.com/kiali/kiali/pull/9192


jshaughn (approved):
LG. It could be more verbose about when you would want to change it from the default value. But it's fine as-is.

PR #951, opened 2026-02-16
update docs regarding the root_namespace auto-detection

### LINKED ISSUE #9236: remove root_namespace section in docs
root_namespace is no longer used - we need to update the docs accordingly.

See: https://kiali.io/docs/configuration/istio/#root-namespace

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/9236

I think we should backport this to `current` branch.

netlify: https://deploy-preview-951--kiali.netlify.app/docs/configuration/istio/#root-namespace

Side-fix: The videos shortcode was failing CI builds when the external feed2json service
returned errors instead of valid YouTube data. The template tried to call Hugo's
'first' function on nil data, causing "both limit and seq must be provided" error.
Added defensive checks to validate JSON response and graceful fallback when videos
can't be loaded, preventing external service outages from breaking site builds.

### DIFF
--- a/content/en/docs/Configuration/istio.md
+++ b/content/en/docs/Configuration/istio.md
@@ -45,15 +45,17 @@ page](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/) and
 search for "rootNamespace".
 
 Kiali uses the root namespace for some of the validations of Istio resources.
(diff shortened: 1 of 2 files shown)


jshaughn (approved):
Good fixes! That shortcode thing is helpful.

PR #952, opened 2026-02-17
[current] update docs regarding the root_namespace auto-detection

jmazzitelli:
I think we should put this into current right now since the root_namespace docs are wrong as of the current release. This is a backport of:
* https://github.com/kiali/kiali.io/pull/951


jshaughn (approved):
Good idea.

PR #953, opened 2026-02-17
Fix staging branch baseURL

ferhoyos:
The release workflow failed on February 16th with:

```
nothing to commit, working tree cleanError: Process completed with exit code 1.
``` 

The staging branch had baseURL = "https://kiali.io" instead of https://staging.kiali.io. When the release script reset current to staging and tried to change the baseURL, there was nothing to change, causing git commit to fail.

This PR sets baseURL = "https://staging.kiali.io" for the staging branch (was incorrectly set to production URL).

Additionally, I have added some changes in the release script:
- Fail with clear error if staging branch has production baseURL (indicates misconfiguration)
- Skip version branch creation if it already exists on remote (for safe release re-runs)
- Skip params.versions addition if version already exists in config.toml (for safe release re-runs)

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1,4 +1,4 @@
-baseURL = "https://kiali.io"
+baseURL = "https://staging.kiali.io"
 title = "Kiali"
 
 enableRobotsTXT = true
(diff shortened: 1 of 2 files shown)


jshaughn (approved):
Oh, good catch! I fixed things a bit but didn't see the root cause.

PR #959, opened 2026-03-09
Replace third-party feed converters in blog and videos shortcodes

jshaughn:
We are experiencing a lot of failed Netlify builds. I think it has to do with the free converters we were using to convert the RSS feeds into something consumable for Hugo. This PR removes that tooling in favor of native handling. It's sort of yucky looking but seems to work.

https://deploy-preview-959--kiali.netlify.app/#td-block-3

This should get backported to `current` and `v2.23`, neither of which are currently deployed successfully.


jmazzitelli (approved):
this sure looks dandy - I'm sure it's awesome.

PR #965, opened 2026-04-16
Document require_namespace_get feature flag for multi-tenant RBAC

jshaughn:
Part-of: https://github.com/kiali/kiali/issues/7125

## Summary
- Adds a new subsection under "Granting access to namespaces" in the RBAC documentation
- Documents the `kiali_feature_flags.authz.require_namespace_get` option
- Explains the multi-tenant use case where LIST is granted broadly but GET is restricted per namespace
- Includes notes on performance implications and caching behavior

https://deploy-preview-965--kiali.netlify.app/docs/configuration/rbac/#granting-access-to-namespaces
https://deploy-preview-965--kiali.netlify.app/docs/configuration/rbac/#multi-tenant-environments-and-require_namespace_get

Made with [Cursor](https://cursor.com)


jmazzitelli (changes requested):
see comment

jmazzitelli:
Nice addition — the new section itself reads well. However, I think there are a few other places in the docs that should reference this new section (even if just a brief note with a link). Here's what I found:

---

**`Configuration/namespace-management.md`**

The info alert at the top of this page carefully distinguishes between "accessible" namespaces (what the Kiali Server can see) and "visible" namespaces (what the end user can see). But it only explains visibility in terms of discovery selectors and `cluster_wide_access`. It never mentions that per-user RBAC (GET vs LIST) further narrows what's visible. A reader could walk away thinking discovery selectors are the *only* knob for filtering user-visible namespaces.

Suggestion: add a sentence to that alert (or a new alert right after it) noting that user visibility is also governed by RBAC rules — with a link to the new [Multi-tenant environments and `require_namespace_get`](https://kiali.io/docs/configuration/rbac/#multi-tenant-environments-and-require_namespace_get) subsection.

---

**`FAQ/general.md` — "Why isn't my namespace in the Namespace Selection dropdown?"**

This FAQ lists reasons a namespace might be missing: discovery selectors, `cluster_wide_access`, operator reconciliation timing, and caching. It never mentions user RBAC as a reason. With `require_namespace_get` enabled, a namespace could be accessible to the Kiali Server yet invisible to a specific user because that user lacks GET permission. This FAQ is a natural troubleshooting entry point.

Suggestion: add a sentence mentioning that RBAC rules (and specifically `require_namespace_get`) can also cause a namespace to be absent from the dropdown, with a link to the new subsection.

---

**`Configuration/multi-cluster/_index.md`** (step 4, "Configure user access")

Step 4 says _"Kiali will check the user's access to each configured cluster's namespace before showing the user any resources from that namespace"_ but doesn't elaborate on how that check works. In multi-cluster + multi-tenant setups, `require_namespace_get` is especially relevant.

Suggestion: add a brief note or link to the new RBAC subsection so readers configuring multi-cluster know about the stricter GET enforcement option.

---

**`Configuration/authentication/openid.md`** (RBAC / RoleBinding section)

This page already links to the RBAC page and shows a `RoleBinding` example granting `get` on namespaces. Since OpenID is the primary non-OpenShift auth strategy for multi-cluster (and multi-tenant environments often use OIDC), a sentence noting that `require_namespace_get` can enforce strict per-namespace visibility — even when LIST is granted broadly — would be valuable near the existing RBAC link or the RoleBinding example.

---

These could all go in this same PR or a follow-up — just flagging them so they don't get missed.

jshaughn:
@jmazzitelli I added all of the suggested touch-points. Please re-review.

jmazzitelli (approved):
I approve. But double-check all those links - I saw one not going to the sub-section and only went to the top of the namespace management page - I'm not sure if it was a browser problem or not.

PR #969, opened 2026-05-08
Indicate how to give access to ztunnel ns

### LINKED ISSUE #9665: [docs] indicate link to say how to give access to ztunnel namespace for ambient features
To enable Kiali to work with Ambient, the docs say:

> For Kiali to detect Ambient, it needs to have access to the namespace where ztunnel is deployed. This is usually the Istio control plane namespace (often istio-system), but on platforms such as OpenShift, it may differ.

See: https://kiali.io/docs/features/ambient/#control-plane-ambient-mesh

But it isn't clear HOW to give Kiali that access.

We should probably provide a link to the namespace management page since that explains how to give Kiali access to specific namespaces in the cluster. So something like this:

> For Kiali to detect Ambient, it needs to [have access to the namespace](https://kiali.io/docs/configuration/namespace-management/) where ztunnel is deployed. This is usually the Istio control plane namespace (often istio-system), but on platforms such as OpenShift, it may differ.

Or maybe we make it more explicit:

> For Kiali to detect Ambient, it needs to have access to the namespace where ztunnel is deployed. This is usually the Istio control plane namespace (often istio-system), but on platforms such as OpenShift, it may differ. To learn how to give Kiali access to specific namespaces see [Namespace Management](https://kiali.io/docs/configuration/namespace-management/).

josunect:
Fixes https://github.com/kiali/kiali/issues/9665

Update: https://deploy-preview-969--kiali.netlify.app/docs/features/ambient/#control-plane-ambient-mesh

### DIFF
--- a/content/en/docs/Features/ambient.md
+++ b/content/en/docs/Features/ambient.md
@@ -26,7 +26,7 @@ This badge indicates that Kiali has detected a `ztunnel` (the L4 component for A
 ![Ambient Control Plane](/images/documentation/features/ambient/ambient-control-plane.png)
 
 {{% alert color="warning" %}}
-For Kiali to detect Ambient, it needs to have access to the namespace where `ztunnel` is deployed. This is usually the Istio control plane namespace (often `istio-system`), but on platforms such as OpenShift, it may differ.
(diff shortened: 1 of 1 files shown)


jmazzitelli (approved):
LG!

PR #976, opened 2026-05-28
update workflows

jmazzitelli:
This PR hardens the GitHub Actions workflows against several known vulnerability classes.

**Supply chain risk (unpinned actions):** All third-party actions were using floating version tags (e.g. `@v4`). These are mutable references — a compromised action maintainer can silently retag them to malicious code that executes in CI with access to repository secrets. All action references have been pinned to their full immutable commit SHA, with the version tag retained as a comment.

**Missing `permissions:` blocks:** Neither the release nor the weekly workflow declared explicit token permissions, meaning they inherited the repository default (typically broad write access). Both workflows now declare `permissions: read-all` at the top level, with `contents: write` granted only to the jobs that actually push commits, deploy to `gh-pages`, or push release branches.

Generated-by: Claude

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out code
-        uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


jshaughn (approved):
Should be fine.

PR #992, opened 2026-07-10
Add opt-in OpenShift impersonation mode for multi-cluster auth

### LINKED ISSUE #10038: OpenShift Impersonation for Multi-Cluster Auth
## Problem

Kiali's multi-cluster support on OpenShift currently requires users to individually authenticate to each remote cluster via the Kiali UI's user profile dropdown. Each cluster login triggers a separate OpenShift OAuth redirect flow. This design is acceptable for small deployments (2-3 clusters) but becomes completely impractical at scale. An organization with 10, 50, or 100+ clusters cannot reasonably ask users to manually log into each one — the experience is so laborious that it effectively renders multi-cluster Kiali unusable for large fleets.

The root cause is that OpenShift OAuth does not provide SSO across clusters. Each cluster has its own OAuth server, its own tokens, and its own session. Kiali faithfully models this by maintaining per-cluster session cookies and requiring per-cluster authentication.

## Scope: OpenShift Only

Impersonation mode is scoped to the openshift auth strategy. Non-OpenShift environments already solve the multi-cluster single-login problem via the openid strategy with a shared OIDC provider — a single OIDC token is valid across all clusters, so there is no per-cluster login burden. Impersonation addresses the gap specific to OpenShift, where each cluster has its own independent OAuth server and there is no built-in cross-cluster SSO.

## Proposal

Add an opt-in impersonation mode for the OpenShift auth strategy. When enabled, the user authenticates once to the home cluster (where Kiali is deployed) via the normal OpenShift OAuth flow. For all cluster API calls (including the home cluster), Kiali then uses its own Service Account credentials (already provisioned on remote clusters via the existing remote cluster secret mechanism) and attaches Kubernetes impersonation headers (Impersonate-User, Impersonate-Group) carrying the authenticated user's identity. The remote cluster's API server authenticates the Kiali SA, verifies it has permission to impersonate, and then evaluates RBAC against the impersonated user — preserving per-user access control without requiring per-user tokens.

This reduces the user login experience from O(n) to O(1) regardless of cluster count, while maintaining the same security boundary: each user can only see and do what their RBAC on each remote cluster permits.

The feature is:

* Opt-in via a new Kiali CR configuration field (`spec.auth.openshift.impersonation.enabled`)
* OpenShift-only (non-OpenShift environments already solve this via the OpenID strategy with a shared OIDC provider)
* Backward-compatible (existing per-cluster OAuth flow remains the default)
* Security-conscious (impersonation privileges are only granted when explicitly opted in; the impersonated identity is derived server-side from the verified home cluster session, never from user-supplied input)

Note: For architectural consistency, impersonation is applied to all clusters including the home cluster. The user's OAuth token is used only for the one-time identity verification; all subsequent API calls use SA + impersonation.

PRs:
* server:  https://github.com/kiali/kiali/pull/10039
* operator: https://github.com/kiali/kiali-operator/pull/1078
* helm-charts: https://github.com/kiali/helm-charts/pull/426
* kiali.io docs: https://github.com/kiali/kiali.io/pull/992

jmazzitelli:
Fixes: https://github.com/kiali/kiali/issues/10038

When spec.auth.openshift.impersonation.enabled is true, users authenticate once to the home cluster and Kiali uses its SA credentials with Kubernetes impersonation headers for all cluster API calls. This eliminates the need for per-cluster OAuth login flows that don't scale beyond a few clusters.

Generated-by: Cursor

Netlify links:
* https://deploy-preview-992--kiali.netlify.app/docs/configuration/authentication/openshift/#impersonation-mode-alternative
* (see step 4 here) https://deploy-preview-992--kiali.netlify.app/docs/configuration/multi-cluster/#setup
* https://deploy-preview-992--kiali.netlify.app/docs/tutorials/ossm-multicluster/ossm-acm-multi-primary/#7-future-impersonation-mode-will-eliminate-per-cluster-login


content/en/docs/Tutorials/ossm-multicluster/ossm-acm-multi-primary.md
```diff
@@ -1483,3 +1483,7 @@ client claims to be in cluster "spoke-two", but we only know about local cluster
 
 When logging into `spoke-two` through Kiali's multi-cluster UI, Kiali redirects to `spoke-two`'s OpenShift OAuth endpoint. The redirect URI must be reachable from the user's browser. If `spoke-two`'s OAuth route is on a different domain, ensure the redirect back to the Kiali URL is reachable.
 
+### 7. Future: Impersonation Mode Will Eliminate Per-Cluster Login
```
jshaughn:
```suggestion
### 7. Impersonation Mode Eliminates Per-Cluster Login
```

jshaughn (commented):
LG, I don't think we need to call impersonation a "Future".

content/en/docs/Tutorials/ossm-multicluster/ossm-acm-multi-primary.md
jmazzitelli: Remember, this is OSSM specific documentation (we  install the operators found in the redhat-operators OLM registry, pre-requisites is OSSM 3.4) - that's why its listed as Future - this will only be available when we ship the next minor-release of OSSM (3.5) which presumably will have this feature. It currently is not in OSSM 3.4

jmazzitelli: I removed the verbiage talking about "future" - instead I just added the blurb that this feature "will be in OSSM 3.5/Kiali" but the rest of the text does not talk about it in the future tense. See https://deploy-preview-992--kiali.netlify.app/docs/tutorials/ossm-multicluster/ossm-acm-multi-primary/#7-impersonation-mode-eliminates-per-cluster-login

jmazzitelli:
We can ignore the build error because the error:

>internally linking to ossm-multicluster-tutorial.sh

just refers to the new script that we added that isn't yet published on kiali.io (it is on staging)

jmazzitelli:
@jshaughn review comments addressed/fixed -- ready for approval?

jshaughn (approved):
Docs look good to me.

PR #1008, opened 2026-08-28
Docs: recording rules and federation for Prometheus tuning

jshaughn:
Closes https://github.com/kiali/kiali/issues/6342
Depends-on: https://github.com/kiali/kiali/pull/10260

## Summary

- Adds a **[Recording Rules and Federation](https://deploy-preview-1008--kiali.netlify.app/docs/configuration/p8s-jaeger-grafana/prometheus/#option-1-recording-rules-and-federation-recommended)** section under [Prometheus Tuning](https://kiali.io/docs/configuration/p8s-jaeger-grafana/prometheus/#prometheus-tuning)
- Documents **production integration**: operators merge reference YAML from `hack/istio/metric-rules/` (`core-recording-rules.yml`, `core-federation-match.yml`, and optional dashboard/Kiali tiers) into their **existing** edge and production Prometheus configuration (`rule_files`, ConfigMap mounts, `PrometheusRule` CRs, Helm values, etc.)
- Separates **production reference files** (repo root under `metric-rules/`) from **`demo/`** scripts and sample deployments (lab/CI only—not for production clusters)
- Includes a production operator checklist and a demo walkthrough (`demo/install.sh`) to validate the pattern locally

Companion implementation/reference PR: kiali/kiali (`issue_6342`)

## Test plan

- [ ] Review doc structure and production vs demo guidance
- [ ] Verify Hugo links (Perses page, FAQ required metrics anchor, production-reference-files anchor)
- [ ] Preview on Netlify deploy

Made with [Cursor](https://cursor.com)

PR #1011, opened 2026-09-10
setting up federated metrics in ACM docs

jmazzitelli:
this goes on top of https://github.com/kiali/kiali.io/pull/1008

netlify:
* https://deploy-preview-1011--kiali.netlify.app/docs/configuration/multi-cluster/acm-observability/
* https://deploy-preview-1011--kiali.netlify.app/docs/tutorials/ossm-multicluster/


jmazzitelli:
I proofread this and it looks correct to me now.

jmazzitelli:
This is close to done - the tutorial is the last I am reviewing and guide 1 installed fine with the script.
need to spin up 3 clusters and test it all - that's next on my list to do

content/en/docs/Configuration/multi-cluster/acm-observability.md
```diff
@@ -30,8 +32,8 @@ Kiali can query these aggregated metrics either through ACM's external Observato
   - Control plane (istiod in istio-system)
   - Ztunnel (in ztunnel namespace, for L4 metrics in Ambient mode)
   - Waypoint proxies (in application namespaces, for L7 metrics in Ambient mode)
-- **Metrics Allowlist ConfigMaps**: Define which metrics ACM should collect
-- **Metrics Collector**: Runs on each managed cluster and pushes its Prometheus metrics to the hub cluster's Thanos every 5 minutes (default)
+- **One MCOA PrometheusRule per target namespace**: Aggregates that namespace's raw per-proxy traffic series into `workload:istio_*` series on UWM
```
jshaughn: This is odd, you need a rule per namespace, and need to update them any time a namespaces is added or removed?

content/en/docs/Configuration/multi-cluster/acm-observability.md
```diff
@@ -53,8 +55,9 @@ There are two independent flows:
 **Ingestion (managed cluster → hub):**
 1. **Istio data plane components** (sidecars, ztunnel, or waypoint proxies) expose metrics at `:15020/stats/prometheus`.
 2. **User Workload Monitoring Prometheus** scrapes those metrics (typically every 30s).
-3. The **ACM observability collector/agent** on the managed cluster reads from Prometheus and ships metrics to the hub (typically every 5 minutes).
-4. The hub stores them in **Thanos Receive/Store** and serves them through **Thanos Query Frontend**.
+3. UWM evaluates recording rules that remove per-proxy cardinality and produce `workload:istio_*` series.
```
jshaughn: Maybe "per-proxy" isn't quite right, maybe "per-pod" or just "scrape-level"?

content/en/docs/Configuration/multi-cluster/acm-observability.md
```diff
@@ -234,55 +249,196 @@ Create a PodMonitor in **each namespace with a waypoint**. Waypoint pods also ex
...
+{{% /alert %}}
+
+### 5. Aggregate and federate Istio metrics
+
+Use UWM as the **Edge Prometheus** described in [Recording Rules and Federation]({{< relref "../p8s-jaeger-grafana/prometheus" >}}#option-1-recording-rules-and-federation-recommended). MCOA's Prometheus Agent is the federation and remote-write layer, and ACM Observatorium/Thanos is the long-retention federated backend that Kiali queries.
+
+Create the following resources on the hub in namespace `open-cluster-management-observability`. Repeat the `PrometheusRule` and platform `ScrapeConfig` for **each** target namespace whose Istio traffic or platform CPU and memory Kiali should display (normally the control plane namespace and every application namespace). The target namespace must exist on every managed cluster selected by the placement.
```
jshaughn: We may need to look more closely at this. It seems like a big ask for the hub to have to explicitly list all mesh namespaces. There could be 100s, maybe more!

content/en/docs/Configuration/multi-cluster/acm-observability.md
jmazzitelli: this is due to default tenancy rules in OpenShift UWM. I'm going to see if its possible to override that and thus have only one PrometheusRule per cluster.

when enabling UWM, we tell it we want a namespace without (namespace) label enforcement:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: user-workload-monitoring-config
  namespace: openshift-user-workload-monitoring
data:
  config.yaml: |
    namespacesWithoutLabelEnforcement:
    - mesh-observability
```

The thinking is maybe this means we can create a PromethesuRule in that mesh-observability namespace and its rules can query metrics from multiple namespaces.

I will experiment with this.

content/en/docs/Configuration/multi-cluster/acm-observability.md
jmazzitelli: changed to: `UWM evaluates recording rules that reduce cardinality and produce `workload:istio_*` series.`

content/en/docs/Configuration/multi-cluster/acm-observability.md
jmazzitelli: I believe we can just create one ScrapeConfig for all platform metrics (regardless of namespace). Cardinality really isn't a concern here. The number of timeseries would be a concern , especially if the majority are not mesh related, but for simplicity sake, we'll just slurp them all in even if we'll be getting more metrics than we really need. I'll look into changing this.

content/en/docs/Configuration/multi-cluster/acm-observability.md
```diff
@@ -53,8 +55,9 @@ There are two independent flows:
 **Ingestion (managed cluster → hub):**
 1. **Istio data plane components** (sidecars, ztunnel, or waypoint proxies) expose metrics at `:15020/stats/prometheus`.
 2. **User Workload Monitoring Prometheus** scrapes those metrics (typically every 30s).
-3. The **ACM observability collector/agent** on the managed cluster reads from Prometheus and ships metrics to the hub (typically every 5 minutes).
-4. The hub stores them in **Thanos Receive/Store** and serves them through **Thanos Query Frontend**.
+3. UWM evaluates the cross-namespace recording rule in the dedicated `mesh-observability` namespace and produces `workload:istio_*` series.
+4. The **MCOA user-workload Prometheus Agent** federates selected series through `/federate`, relabels `workload:istio_*` back to `istio_*`, and remote-writes them to the hub (every 5 minutes by default).
```
jshaughn, line 59: The default of 5' is fairly slow. I wonder if we should recommend anything more frequent. With 5m I think Kiali will have a minimum Duration setting of 10m.

content/en/docs/Configuration/multi-cluster/acm-observability.md
```diff
@@ -234,55 +249,210 @@ Create a PodMonitor in **each namespace with a waypoint**. Waypoint pods also ex
...
+        namespace: open-cluster-management-observability
+      - group: monitoring.rhobs
+        resource: scrapeconfigs
+        name: kiali-istio-platform-federation
+        namespace: open-cluster-management-observability
+```
 
 {{% alert color="info" %}}
-**Ambient Mode**: The same allowlist works for all Istio data plane components. However, ztunnel only produces TCP metrics (`istio_tcp_*`), so HTTP metrics in the allowlist will have no data from ztunnel. Waypoints produce both TCP and HTTP metrics, same as sidecars.
+For a repeatable development installation, the Kiali repository's `hack/configure-acm-mcoa.sh install` helper creates these same objects and placement references without changing the current kubeconfig context. It is a convenience, not a prerequisite for this configuration.
```
jshaughn, line 450: I wonder if we should create a `hack/acm` folder to group all of the acm hack scripts

content/en/docs/Configuration/multi-cluster/acm-observability.md
```diff
@@ -234,55 +249,210 @@ Create a PodMonitor in **each namespace with a waypoint**. Waypoint pods also ex
...
+{{% /alert %}}
+
+{{% alert color="warning" %}}
+The `mesh-observability` namespace must exist on every managed cluster selected by the placement before MCOA propagates its recording rule. If clusters use different application namespace layouts, create the appropriate `ServiceMonitor` and `PodMonitor` resources locally on each cluster. The same MCOA aggregation rule can still be propagated to `mesh-observability` on every cluster.
```
jshaughn, line 454: I wonder if the OSSM ACM add-on should create this `mesh-observability` dir.

content/en/docs/Configuration/multi-cluster/acm-observability.md
```diff
@@ -452,7 +622,7 @@ spec:
       # Enable Thanos proxy mode
       thanos_proxy:
         enabled: true
-        retention_period: "14d"
+        retention_period: "365d"
```
jshaughn, line 627: this isn't really necessary. We should probably just leave it at 2 weeks. Same below...

content/en/docs/Configuration/multi-cluster/acm-observability.md
```diff
@@ -234,55 +249,210 @@ Create a PodMonitor in **each namespace with a waypoint**. Waypoint pods also ex
...
+  namespace: open-cluster-management-observability
+  annotations:
+    observability.open-cluster-management.io/target-namespace: mesh-observability
+  labels:
+    app.kubernetes.io/component: user-workload-metrics-collector
+    openshift.io/prometheus-rule-evaluation-scope: leaf-prometheus
+spec:
+  groups:
+  - name: istio.workload-aggregation
+    interval: 30s
```
jshaughn, line 319: btw, this should not be more frequent than the actual raw scrape, we should also be recommended a 30s scrape for UWM (the edge prom), if it's not at least that frequent already.

jshaughn (commented):
This looks sane, good stuff. A few comments but close.

content/en/docs/Configuration/multi-cluster/acm-observability.md
jmazzitelli: this is done. one promrule per cluster now.

content/en/docs/Configuration/multi-cluster/acm-observability.md
jmazzitelli: this is done.

content/en/docs/Configuration/multi-cluster/acm-observability.md
jmazzitelli: 5m is the OpenShift default. Since we do not have any performance data at scale providing evidence otherwise, I suggest we leave our recommendation to be the same.

content/en/docs/Configuration/multi-cluster/acm-observability.md
jmazzitelli: I have "the same but different" hack scripts and docs in three different repos (kiali/kiali, openshift-servicemesh-plugin, and kiali.io). They are different enough that I didn't yet try to combine things. We could look into consolidating all our ACM related scripts and docs, probably here in this kiali.io repo, but that should be followup work.

content/en/docs/Configuration/multi-cluster/acm-observability.md
jmazzitelli: The PR in openshift-servicemesh-plugin repo that I'm working on, along with this kiali.io PR, has this change, too. See https://github.com/kiali/openshift-servicemesh-plugin/pull/862

But that's just hack scripts. There are no official docs for that stuff yet. When we do write them, yeah, it should follow the same kind of docs as this kiali.io docs.

content/en/docs/Configuration/multi-cluster/acm-observability.md
jshaughn: OK, we probably don't want to rock the boat. If a Kiali user wants a min duration < 10m they will have to lower this themselves.

content/en/docs/Configuration/multi-cluster/acm-observability.md
jshaughn: fine. I just feel like the `kiali/hack` folder is getting big.

content/en/docs/Configuration/multi-cluster/acm-observability.md
jmazzitelli: This is just tracking the default for the hub Observatorium backend storage. All that setting does is tell Kiali, "_This is what the backend Thanos retention is_". And that value is simply the default (again, the OpenShift default).

I don't think we want to recommend anything other than the OpenShift defaults because we need to remember this ACM Hub backend storage is not just used for Kiali or OSSM but for EVERYTHING in the user's multi-cluster environment. We shouldn't tell the user to do anything special just for us when that recommendation might be a bad one for something else they want to use ACM Thanos for.

For the record, here's the actual Observatorium code - you can see their defaults is 1 year for these retention periods:

* Where the defaults are defined: https://github.com/stolostron/multicluster-observability-operator/blob/release-2.17/operators/multiclusterobservability/pkg/config/config.go#L192-L194
```go
RetentionResolutionRaw = "365d"
RetentionResolution5m  = "365d"
RetentionResolution1h  = "365d"
```
* where [the defaults are set](https://github.com/stolostron/multicluster-observability-operator/blob/release-2.17/operators/multiclusterobservability/controllers/multiclusterobservability/observatorium.go#L967-L986):

```go
	if mco.Spec.AdvancedConfig != nil && mco.Spec.AdvancedConfig.RetentionConfig != nil &&
		mco.Spec.AdvancedConfig.RetentionConfig.RetentionResolutionRaw != "" {
		compactSpec.RetentionResolutionRaw = mco.Spec.AdvancedConfig.RetentionConfig.RetentionResolutionRaw
	} else {
		compactSpec.RetentionResolutionRaw = mcoconfig.RetentionResolutionRaw
	}
... same for the 5m and 1h ...
...
		compactSpec.RetentionResolution5m = mcoconfig.RetentionResolution5m
...
		compactSpec.RetentionResolution1h = mcoconfig.RetentionResolution1h
...
```

jshaughn: Fine.

content/en/docs/Configuration/multi-cluster/acm-observability.md
jshaughn: Do we know what the global scrape is set to for the UWM edge (the actual scrape of the istio raw metrics)?

jmazzitelli: I'm going to go through everything and make sure these examples explicitly set the values so we can see where things are 30s and where things are 5m.

I'm going to read up and get more familiar. ScrapeConfig, PrometheusRule, PodMonitor, and ServiceMonitor all have intervals and we need to make sure they are set properly to work in concert together.

jmazzitelli: > we should also be recommended a 30s scrape for UWM

That is in here.

See the latest commit - I added a section in acm-observability.md (that the tutorials now refer back to, rather than repeatedly regurgitate this info). It explains the scrape intervals.

jshaughn: OK, this new content is great, exactly what we needed to explain to users (and ourselves).

jshaughn (approved):
This looks really good now.

jshaughn (approved):
I don't fully grok the latest changes but I'll take your word for it.

